PR #1, opened 2014-12-24
Win64 crash in asm.js

InternalError503:
Return stack offset before incrementing it for win64 Bug #1071444
P.s Merry Christmas


MrAlex94:
Thanks a lot for the help mate, haven't had time to look into it, you've done me a big favour! 

Happy Holidays :)

Issue #95, opened 2017-01-29
Add config flag to use native tab duplication


Scrxtchy:
As introduced in 89eb998322da3339c6beac1398b283e1f252f415

Can this be made removable with a configuration flag?

PandaCodex:
@Scrxtchy A PR with user controlled preference item has been proposed to Waterfox developer. He needs to review and merge the request once approved.

_Impact to browser is minimal with code maintenance very low._

PR #96, opened 2017-02-01
User controlled duplicate tab menu item.

PandaCodex:
This change allows users to toggle `"browser.tabs.duplicateTab"` preference. When toggled the `Duplicate Tab` menu item toggles.

Many extensions like tab mix plus add these entries with users ending up with multiple entries for the same function. This allows **users choice** in customization of there browser installation.

Linking to current tracked issue https://github.com/MrAlex94/Waterfox/issues/95

### DIFF
--- a/browser/app/profile/firefox.js
+++ b/browser/app/profile/firefox.js
@@ -446,6 +446,7 @@ pref("browser.tabs.loadDivertedInBackground", false);
 pref("browser.tabs.loadBookmarksInBackground", false);
 pref("browser.tabs.tabClipWidth", 140);
 pref("browser.tabs.animate", true);
+pref("browser.tabs.duplicateTab", true);
 #ifdef UNIX_BUT_NOT_MAC
 pref("browser.tabs.drawInTitlebar", false);
 #else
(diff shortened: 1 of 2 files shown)


PandaCodex:
@MrAlex94 is there are reason for `<menuitem id="context_duplicateTab" label="Duplicate Tab" accesskey="D"` hardcoding the strings to the menu item.

When adding entities to localization files, same has to be done with language packages or an entity undefined error will be thrown. Out of the box on standard en-US with the entries added to browser.dtd should work without issue.

Did you want a PR made for this ?

MrAlex94:
Yes, I hard coded them on purpose. For some reason when I added the strings to the locale file, XUL would complain that it couldn't find the label for the tab. I have no idea why and I made sure every character was exactly the same and that there was no strange formatting going on. You can have a look and see if it does the same thing for you as well.

PandaCodex:
@MrAlex94 was able to build without error adding the entities to the browser.dtd then adding
`<menuitem id="context_duplicateTab" label="&duplicateTab.label;" accesskey="&duplicateTab.accesskey;"`

Did you make the change then rebuild on an existing build i.e non clobber. Try a full clobber also check that the en-US language pack is not installed in features or profile.

MrAlex94:
Yep, 100% sure. I'll send you the build when I've got a decent connection again. By all means it should work without problem. Although sometimes Visual Studio has been known to act weird after a while without a restart so that might be the issue.

Issue #99, opened 2017-02-01
[DOS] Crash with language selection


PandaCodex:
Patch https://github.com/MrAlex94/Waterfox/commit/efa1f09c3d1478387538c64afac5a6dd50405349 brought with it a series of issues.

- Using arrow keys for selection causes large CPU spiking
- Using arrow keys causes memory leak from notification spam
- Using arrow keys can crash waterfox
- Using arrow keys creates browser lag creating a **denial of service** for few seconds to few minutes

Currently using the arrow keys to select language causes the above issues, every change made triggers `nsIAlertsService` when scrolling 91 entries that is 91 notifications passed through `nsIAlertsService` this causes a memory leak and can crash the browser.

Suggestion would be to only select the selected item once the drop down menu closes creating a single notification and resolving the above issues.

![dos](https://cloud.githubusercontent.com/assets/25470255/22506041/05ef2a18-e8ca-11e6-9c49-b637c591d948.JPG)

Lastly the field remains empty even after changed, there is no visual feedback to the current default selected language, for usability and readability this is highly suggested.

MrAlex94:
Yes I was rather hasty with the way I implemented this, I'll have to go back and choose a better method. Maybe a drop down box isn't the best way to accomplish this.

PandaCodex:
@MrAlex94 The current PR work for language selection fixes should resolve the above mentioned issues.

PR #100, opened 2017-02-01
Fix crash and dos with language selection

PandaCodex:
This is to address patch efa1f09 brought with it a series of issues and reported issue #99.

-  ~~Using arrow keys for selection causes large CPU spiking~~
-  ~~Using arrow keys causes memory leak from notification spam~~
-  ~~Using arrow keys can crash waterfox~~
-  ~~Using arrow keys creates browser lag creating a denial of service for few seconds to few minutes~~

This proposed change only fires an event when the language selection popup closes, ~~this however does introduce another limitation. When using tab to select control element then when selecting with arrow keys no longer selects the items but on `onkeypress` event with function preventing default and binding to the enter (return key) once selection with arrow keys are made when using tab index to select GUI elements.~~

This change also sets the current default language without triggering `updateLocale` event.

### DIFF
--- a/browser/components/preferences/in-content/main.js
+++ b/browser/components/preferences/in-content/main.js
@@ -87,7 +87,10 @@ var gMainPane = {
     setEventListener("localeSelect", "popuphiding", function () {
       gMainPane.updateLocale();
     });
-
+    setEventListener("localeSelect", "keypress", function (e) {
+      gMainPane._updateLocale(e);
+    });
+	
     if (AppConstants.E10S_TESTING_ONLY) {
       setEventListener("e10sAutoStart", "command",
                        gMainPane.enableE10SChange);
@@ -143,7 +146,13 @@ var gMainPane = {
       alertsService.showAlertNotification("",  "Restart Waterfox", "You'll need to restart Waterfox to see your selected locale.");
     }
   },
-
+  _updateLocale: function (e)
+  {
+    if (e.which == 13 || e.keyCode == 13){
+		  this.updateLocale();
+    }
+  },
+    
   enableE10SChange: function ()
   {
     if (AppConstants.E10S_TESTING_ONLY) {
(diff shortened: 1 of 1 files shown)


browser/components/preferences/in-content/main.js
```diff
@@ -28,7 +28,7 @@ var gMainPane = {
       document.getElementById(aId)
               .addEventListener(aEventType, aCallback.bind(gMainPane));
     }
-    this.updateLocale();
+    this.getDefaultLocal();
```
PandaCodex, line 31: Sets language selector to current locale value without `updateLocale` notification.

browser/components/preferences/in-content/main.js
```diff
@@ -84,7 +84,7 @@ var gMainPane = {
     setEventListener("chooseFolder", "command",
                      gMainPane.chooseFolder);
 
-    setEventListener("localeSelect", "command", function () {
+    setEventListener("localeSelect", "popuphiding", function () {
```
PandaCodex, line 87: Only fires `updateLocale` function when dropdown menu popup closes.

browser/components/preferences/in-content/main.js
```diff
@@ -87,7 +87,10 @@ var gMainPane = {
     setEventListener("localeSelect", "popuphiding", function () {
       gMainPane.updateLocale();
     });
-
+    setEventListener("localeSelect", "keypress", function (e) {
```
PandaCodex: Binds to the `keypress` event, this case when keys are pressed on the focused control

browser/components/preferences/in-content/main.js
```diff
@@ -143,7 +146,13 @@ var gMainPane = {
       alertsService.showAlertNotification("",  "Restart Waterfox", "You'll need to restart Waterfox to see your selected locale.");
     }
   },
-
```
PandaCodex, line 138: Special event function that will call `updateLocale` when the **ENTER (Return Key)** is pressed

MrAlex94:
This looks great! Thanks for the patches so far. I'll make sure to have this added as well.

Issue #98, opened 2017-02-01
Sponsored tiles


PandaCodex:
Sponsored tiles with analytical tracking campaigns still exist in Waterfox despite multiple places saying they have been removed or disabled.

Shared profiles with Firefox is the cause, using own profile system is recommended for how you toggle this browser feature. To share the profile suggestions would be to remove the sponsored tile code from the browser core.

![sponsored_tiles](https://cloud.githubusercontent.com/assets/25470255/22505812/08b5befc-e8c9-11e6-8780-2ff9987e5d6d.JPG)

PandaCodex:
@MrAlex94 did you want a patch to remove this feature to prevent further issue?

MrAlex94:
Yes please, that would be much appreciated. I think I've got an old patch somewhere to completely gut the code but that was a while ago. I'll merge all the patches when I'm not traveling as much, thanks for all the help so far!

MrAlex94:
Oh yes also, in Waterfox 52 I'll be making a separate profile location and import users existing profile, so should alleviate this from happening in the future as well.

never93489:
> separate profile location

[hooray emoji]

PandaCodex:
Please see PR work https://github.com/MrAlex94/Waterfox/pull/110 and review the changes for the above issue.

MrAlex94:
Cool. Thanks for all the patches. I'll have a look through them all. Some __may__ not be needed as a lot of components don't get built at all such as the health report, so even if someone has Firefox resetting the configs when they use Waterfox nothing should be happening.

PandaCodex:
> Cool. Thanks for all the patches. I'll have a look through them all. Some may not be needed as a lot of components don't get built at all such as the health report, so even if someone has Firefox resetting the configs when they use Waterfox nothing should be happening.

None of the patches submitted do anything for Firefox Health Report, mind you Firefox Health Report is still enabled in Waterfox looking back through older builds it was enabled too, would you like a patch to disable Firefox Health Report. Currently if shared profile with Firefox Waterfox will submit telemetry data because it gets enabled by Firefox also the data from Waterfox gets submitted by Firefox so technically Waterfox is still collecting and submitting user telemetry data.

I can have a patch later tonight if urgent for Waterfox 52.0 release let me know.

Edit: added images

![hreport](https://cloud.githubusercontent.com/assets/25470255/23641260/269b226a-033e-11e7-88c4-92c90b2524cc.JPG)
[...]

PR #110, opened 2017-03-04
Remove Sponsored Tiles

PandaCodex:
This change removes sponsored tiles completely from Waterfox core. This change will keep the sponsored
tiles removed if shared profile with Firefox.

This change is **medium impact** with slightly more changed then cosmetics while still safe **_its suggested not to merge into the master branch until tested_** to ensure the patch is complete without issues.

This change needs testing for breakages of the about:newtab page styling and grid javascript. Please test and let me know if any issues thank you.

This is in reference to issue https://github.com/MrAlex94/Waterfox/issues/98


PandaCodex:
This could have merged cleanly but now its dirty and will have to be fixed to merge. Time is valuable so someone else can take this i put enough time in it. Sponsored tiles are still in Waterfox 52.0- 52.0.1

PR #115, opened 2017-03-04
Add plugin check back in about:addons

PandaCodex:
This is an optional change, looking at your users request on overclockers many were asking for this back, this change is low impact just a xul link.

If you do take this change did you want the localized strings ?


MrAlex94:
Thank you for this.

PandaCodex:
No problems

PR #122, opened 2017-03-15
Add dont-override-general-useragent-locale.patch for Waterfox Linux

hawkeye116477:
This patch prevents overwriting general.useragent.locale, so if I set intl.locale.matchOS to true, then general.useragent.locale is pl (my default system language), without this patch this value will be overwritten by en. 

http://bazaar.launchpad.net/%7Emozillateam/firefox/firefox.xenial/view/head:/debian/patches/dont-override-general-useragent-locale.patch


MrAlex94:
Sorry haven't time to look at this yet! I should ideally only have this pref set for Windows as I need to force Silverlight for Netflix/Amazon so I'll edit the patch and then pull it.

PR #155, opened 2017-05-29
Update of Polish translation

hawkeye116477:


MrAlex94:
Sorry for taking so long to check this. Will do it now.

PR #165, opened 2017-06-16
Small fix for dont-override-general-useragent-locale.patch

hawkeye116477:
I removed two lines, because these lines are just not needed.


hawkeye116477:
@MrAlex94

MrAlex94 (approved):

PR #185, opened 2017-08-17
Add missing icons

hawkeye116477:


MrAlex94:
Thanks for the patch! I realised the mistake, there are some other issues with the unofficial branding that I don't think Mozilla have seen (language packs undefined parts), so will be copying over files from official branding :-)

hawkeye116477:
@MrAlex94 Ok

PR #191, opened 2017-08-25
Even Slack is too Slack to support Waterfox!

Jimmyson:
But seriously, we need to re-think this user-agent thing!


metux:
Slack is slow on all browsers I've tested so far.
Does user agent tweaking bring any notable improvement ?

Perhaps we should consider moving out that stuff to an external component, which takes these site-specific settings from an external data source, which operators / distros can easily maintain ...

Scrxtchy:
@metux makes the website work
Here's what it looks like without a useragent override
Slack don't keep a list of what they do and don't support either, and currently the compatibility between waterfox and firefox is pretty insignificant that they technically do support it, but their back-end won't care.
They don't even support Opera, which is a pretty big change from Chromium for them to handle

Pretty funny since people are paying quite a lot of dollars on this platform, whereas FOSS end up with wider compatibility 

image

MrAlex94:
Added :-)

PR #199, opened 2017-08-31
Add PulseAudio support

### LINKED ISSUE #184: PulseAudio Support
It appears that Waterfox for Linux is compiled without PulseAudio support.
Unfortunately PulseAudio's ALSA emulation is problematic on my system, causing occasional buffer underruns.

It would be nice if Waterfox binaries were compiled with both PulseAudio and ALSA support, meaning that supported platforms can use PA without having to deal with a buggy emulation layer, while still having working audio for ALSA only systems.

hawkeye116477:
Fixes #184

### DIFF
--- a/.mozconfig
+++ b/.mozconfig
@@ -20,7 +20,7 @@ export CXX=clang++
 ac_add_options --enable-optimize
 ac_add_options --target=x86_64-pc-linux-gnu
 ac_add_options --with-ccache=ccache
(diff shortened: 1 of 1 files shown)


hawkeye116477:
@MrAlex94

MrAlex94:
Thanks, meant to resolve this after the hubbub back a few versions ago.

PR #208, opened 2017-09-12
Update ecosia.xml

WagnerGMD:
This time I will give 1 example. Then instead of : 
```
https://www.ecosia.org/search?q=Waterfox&ref=waterfox
```
You will get one clean url : 
```
https://www.ecosia.org/search?q=Waterfox
```


PandaCodex:
@WagnerGMD The `&amp;ref=waterfox` is so alex gets money from searches made and search data sold by ecosia from the data they gather.

WagnerGMD:
> Waterfox is partners with Ecosia, a search engine that plants trees with its generated revenues.

@PandaCodex I understand because indeed it's true (partners). On the moment, I didn't think about and I didn't realize.

PR #214, opened 2017-09-12
Add Qwant

WagnerGMD:
Right now, [Qwant](https://www.qwant.com) is missing. It will provide another choice as search engine.
You can check (and download) these files :
- [Qwant](https://www.qwant.com/opensearch.xml)
- [Qwant Lite](https://lite.qwant.com/opensearch.xml)


MrAlex94:
Thanks for the requests @WagnerGMD, I've changed most of the searches now to their OpenSearch equivalent where possible.

PR #284, opened 2017-11-27
Mismerge fixes for 56.0

jbeich:
`git diff FIREFOX_56_0_2_RELEASE..waterfox/master` currently shows a lot of whitespace noise. Most of that is caused by recent 56.0 merge and will get in the way of backporting fixes from Firefox 57+.


MrAlex94:
Thanks, merged.

PR #337, opened 2017-12-28
Use ~/.waterfox instead ~/mozilla for user extensions and plugins

### LINKED ISSUE #327: ~/.waterfox/plugins/ and ~/.waterfox/extensions/
@MrAlex94 do you foresee any situation where a user might have `~/.waterfox/` **and not** the traditional `~/.mozilla/` directory?

If so, then maybe enhance Waterfox to work with: 

- `~/.waterfox/plugins/`

## Background

Whilst troubleshooting the Oracle side of #322 I experimented with `~/.waterfox/plugins/`: 

- with symbolic link `~/.waterfox/plugins/libnpjp2.so` the plugin is not recognised; is not listed at `about:plugins`
- with symbolic link `~/.mozilla/plugins/libnpjp2.so` the plugin is recognised, listed (but not working, as outlined in #322).

PhantomX:
This fixes #327 .

### DIFF
--- a/caps/NullPrincipal.cpp
+++ b/caps/NullPrincipal.cpp
@@ -110,7 +110,7 @@ NullPrincipal::Init(const OriginAttributes& aOriginAttributes, bool aIsFirstPart
   if (aIsFirstParty) {
     // remove the '{}' characters from both ends.
     path.Mid(path, 1, path.Length() - 2);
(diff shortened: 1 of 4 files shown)


MrAlex94:
Excellent, thank you!

PR #348, opened 2018-01-01
Fix default browser

### LINKED ISSUE #343: Waterfox 56 is being incorrectly registered as a Default Application in Windows
The issue I reported via Reddit [Waterfox is being incorrectly registered as a Default Application in Windows](https://www.reddit.com/r/waterfox/comments/3xrupj/waterfox_is_being_incorrectly_registered_as_a/) two years ago which I think was fixed at the time, now exists again in Waterfox 56.

This is the reason why Issue #315 exists as this in effect makes only the most recent application that writes the FirefoxHTML and FirefoxURL registry keys able to be a default browser.

The only browser that should write to those registry keys is Firefox but currently Waterfox 56 does this as does the latest version of Basilisk.

I noticed this issue do to installing Basilisk and not being able to to launch Waterfox as a Default Application anymore.
As this also means Basilisk have the same issue, I'll report it to them too.

jdunn0:
This fixes #343 by changing the relevant registry keys/values to use Waterfox instead of Firefox.
Keys/Values Changed:
* Software\Classes\FirefoxHTML -> Software\Classes\WaterfoxHTML
* Software\Classes\FirefireURL -> Software\Classes\WaterfoxURL
* Software\Clients\StartMenuInternet\WATERFOX.EXE\Capabilities\FileAssociations key values changed to WaterfoxHTML
* Software\Clients\StartMenuInternet\WATERFOX.EXE\Capabilities\URLAssociations key values changed to WaterfoxURL
* Software\Mozilla\Firefox -> Software\Mozilla\Waterfox
Depending on certain conditions, these values will be created in this key: DefaultBrowserOptOut,  OldDefaultBrowserCommand and Uninstalled-${UpdateChannel}.

Though using Software\Mozilla\Waterfox is not ideal as Waterfox is not a Mozilla product, other places in the code use the Software\Mozilla as a base key already so I did this to be consistent.

If someone wants to change all the places that use that key to use Software\WaterfoxCompany or something instead then doing so would be better in a separate pull request.

### DIFF
--- a/browser/components/shell/ShellService.jsm
+++ b/browser/components/shell/ShellService.jsm
@@ -62,10 +62,10 @@ let ShellServiceInternal = {
 
     if (AppConstants.platform == "win") {
       let optOutValue = WindowsRegistry.readRegKey(Ci.nsIWindowsRegKey.ROOT_KEY_CURRENT_USER,
-                                                   "Software\\Mozilla\\Firefox",
+                                                   "Software\\Mozilla\\Waterfox",
                                                    "DefaultBrowserOptOut");
       WindowsRegistry.removeRegKey(Ci.nsIWindowsRegKey.ROOT_KEY_CURRENT_USER,
-                                   "Software\\Mozilla\\Firefox",
+                                   "Software\\Mozilla\\Waterfox",
                                    "DefaultBrowserOptOut");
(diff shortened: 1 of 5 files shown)


MrAlex94:
Thanks for the patch!

Issue #384, opened 2018-01-16
Detailed Info on the issue of being the Default Browser on Windows and what may still need fixing


jdunn0:
There have been a few issues with getting Waterfox 55 and higher to correctly handle being the default browser in Windows.
I have created this Github issue to list how Firefox and Waterfox handle being set as the default browser in Windows, what may still need to be fixed and keep track of previous Github Issues and Reddit Posts on the topic in one place.

On Github those issues are: #315, #343, #363, #383.
On Reddit the posts about this are [TBD].

There has been some progress on this topic and new installs of Waterfox 56.0.3 where no previous version of Waterfox has been installed before, the ability the set the default browser to Waterfox should always work without issue.

Unfortunately, if have a previous version of Waterfox installed before, some older registry keys could be left over that cause issues when trying to set Waterfox as the default browser.

Here is how the Installer Registers as a default browser handler in Windows:
1\. Application Registration information with a specific ClientID is written.
If the installer has administrator privileges, the key is under the HKLM registry hive.
If the installer doesn't have administrator privileges and the Operating System is Windows 8 or higher then the key is under the HKCU registry hive.
If the installer doesn't have administrator privileges and the Operating System is not Windows 8 or higher then the key is not written at all as registering a client under HKCU is only available on Windows 8 or higher.
The key is Software\Clients\StartMenuInternet\\[ClientID].
This key has many sub keys but the Capabilities key is the one used by Windows Vista and higher.
The other sub keys are for displaying in the Start Menu in Windows XP/Vista and the SPAD feature introduced in Windows XP SP1 that is unnecessary starting with Windows Vista and removed in the Windows 10 Fall Creators Update.
The Capabilities key can be located anywhere so a value needs to be created under Software\RegisteredApplications that points to "Software\Clients\StartMenuInternet\\[ClientID]\Capabilities", the value name can either be the full name of the program or [ClientID] for consistency.

In early versions of Firefox and Waterfox, [ClientID] was called FIREFOX.exe and WATERFOX.exe.
In later versions of Firefox when multiple editions of Firefox were made available such as the release edition and the Developer Edition, using FIREFOX.exe was not specific enough so it was changed to $AppRegName-$AppUserModelId where $AppUserModelId is a specific id generated by the installer, the first time it's run (upgrade installs use the already existing $AppUserModelId instead of generating a new one) and $AppRegName is Firefox.
Later versions of Waterfox picked up this change though it's not really needed for Waterfox at the current time.
[...]

Issue #363, opened 2018-01-08
Windows 7 Waterfox 56.0.2 Keeps asking to be set as default browser dispite already being default.


mikeloeven:
After updating to 56.0.2 I keep getting the set as default browser dialogue dispite already being the default browser. Clicking set as default does not resolve the issue and the do not check button is ignored

grahamperrin:
https://github.com/MrAlex94/Waterfox/pull/362

grahamperrin:
Cross reference https://www.reddit.com/r/waterfox/comments/7ov4a3/waterfox_5602_is_now_available_with_important/dsd00am/

Ofunniku:
Current workaround: Go to "Default Programs" and set Waterfox as default there.

mikeloeven:
Waterfox has no entry in default programs. there was a Firefox entry that had the waterfox icon but this did not resolve the issue. downloaded and ran a fresh copy of the installer attempting to repair. also no change. the default programs listing with the waterfox icon listed as firefox seems to indicate that something was not refactored properly

Ofunniku:
![2018-01-09 13_42_56-set default programs](https://user-images.githubusercontent.com/5677089/34706515-3843dc30-f543-11e7-8c9f-39d9aa6f5f1a.png)

Yeah, that's most certainly an issue on your end.

jdunn0:
After seeing this bug, I decided to test installing/uninstalling Waterfox 56.0.2 to see if I could figure out what is happening and it appears that Waterfox 56.0.2 is installing registry keys that use Waterfox instead of Firefox in all but one registry key via my patch in #348.
That one registry key should now be fixed via my patch in #362.

The one registry key is installed in Software\Clients\StartMenuInternet in HKCR if the user is doing a per-user install or HKLM or the user is doing a per-machine install.
It appears the installer will pick per-machine install if the user says yes to a UAC prompt and a per-user install if the user says no.

For Firefox and Waterfox, this key will be named either Firefox or Firefox-[AppModelId].
The [AppModelId] part is some ID generated upon install and is used to allow different browser registration entries to appear when multiple editions of Firefox are installed such as the Release version and the Developer version.

I'm not sure how the installer determines when to use Firefox or Firefox-[AppModelId] as the key name.
As long as it's picks the later then the default browser registration done by Waterfox 56.0.2 works just fine and both Firefox and Waterfox will show up in "Set Default Programs".
[...]

PR #396, opened 2018-01-20
Update Installer to remove old registry keys/values that use "Firefox"

jdunn0:
This is a better patch then the one I did in pull request #372 that should solve the problem of duplicate Waterfox entries in Set Default Programs described in #384.

After spending more time looking to the installer code, I found the following:
1. I didn't need to use ClearErrors as if appears that any code that checks for errors already will call that first before an operation that it needs to look at the error results of.
2. The existing installer code already uses DeleteRegKey and DeleteRegKey to handle older keys that may not exist so it is the correct way to deal with this issue.
3. When removing old registry keys, it appears better to just delete them from both the HKLM and HKCU keys as you can't be sure that the root key the install currently uses is the only place they might be.
4. Put the removal of the browser registration keys in the RemoveDeprecatedKeys macro as that marco is already there for the purpose of removing old keys/values that aren't used anymore.
5. Add code to remove the old values added by the install that adds the app model id to the registry.
Functionally, this should make no difference but it's probably best to remove values the installer doesn't use anymore that would only be used by older Waterfox versions.

I'm a lot more confident in this code then the pull request I did before to clean up old browser registration keys.
Hopefully after this patch, there will no longer be any problems when someone wants to set Waterfox as their default browser.

### DIFF
--- a/browser/installer/windows/nsis/shared.nsh
+++ b/browser/installer/windows/nsis/shared.nsh
@@ -1000,6 +1000,12 @@
   ; Remove protocol handler registry keys added by the MS shim
   DeleteRegKey HKLM "Software\Classes\Waterfox.URL"
   DeleteRegKey HKCU "Software\Classes\Waterfox.URL"
+
+  ; Remove old browser registration with "Firefox" prefix from registry
+  DeleteRegKey HKLM "Software\Clients\StartMenuInternet\Firefox-$AppUserModelID"
+  DeleteRegValue HKLM "Software\RegisteredApplications" "Firefox-$AppUserModelID"
+  DeleteRegKey HKCU "Software\Clients\StartMenuInternet\Firefox-$AppUserModelID"
+  DeleteRegValue HKCU "Software\RegisteredApplications" "Firefox-$AppUserModelID"
 !macroend
 !define RemoveDeprecatedKeys "!insertmacro RemoveDeprecatedKeys"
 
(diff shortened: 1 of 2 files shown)


browser/installer/windows/nsis/shared.nsh
```diff
@@ -1000,6 +1000,12 @@
   ; Remove protocol handler registry keys added by the MS shim
   DeleteRegKey HKLM "Software\Classes\Waterfox.URL"
   DeleteRegKey HKCU "Software\Classes\Waterfox.URL"
+
```
PandaCodex, line 1003: Will this affect default Firefox installation if Firefox is default browser when installing Waterfox?

jdunn0: The part "Remove protocol handler registry keys added by the MS shim" did reference Firefox until I changed it in #348 to be Waterfox. I'm not what the "MS shim" refers to but figured it should be changed.

On deleting the keys/values of Firefox-$AppUserModelID, I checked the installer code and it turns out that $AppUserModelID is a hash value calculated from the installation path.

Because of that, the name was unique and did not conflict with Firefox but I changed it in pull request #362 as I found the the display name for the ProgIDs is "${AppRegName} HTML Document" so associations ended up being described as "Firefox HTML Document" which would be confusing and I thought there might be other parts of the code that use ${AppRegName} that might be fixed by the change also.
Later when looking at issue #363, it appeared the change may fix that issue too (I'm not really sure though, I'm still looking in to the issue).

Thus Waterfox 56.0.3 uses the [ClientID] of Waterfox-$AppUserModelID and doesn't know that previous versions used Firefox-$AppUserModelID so on upgrade, there would be two ClientIDs for Waterfox which causes two entries to appear in Set Default Programs for Waterfox so this patch will remove that key so one one Waterfox entry appears in Set Default Programs.

So in summery: This keys existence has no affect on Firefox, as the installation path would need to be the same for the $AppUserModelID parts to match and installing Waterfox in the same directory as an install of Firefox would cause many more problems then this.

PR #443, opened 2018-02-18
add missing default pref

laniakea64:
Fixes error when right-clicking a tab


MrAlex94:
Good catch, thank you.

Issue #380, opened 2018-01-14
e10s: multiprocess enabled with incompatible add-ons after refreshing waterfox 56.0.3


grahamperrin:
From [Refresh Firefox - reset add-ons and settings | Firefox Help](https://support.mozilla.org/kb/refresh-firefox-reset-add-ons-and-settings)

> … If you're having problems …, refreshing it can help. …

With Waterfox 56.0.3: a refresh causes multiprocess to be enabled with add-ons that are not compatible. There's a significant risk of causing problems; not helpful.

AjvarXX:
56.0.3 implemented new option of enabling e10s with unsupported addons and using single process for those unsupported addon.
_Enable e10s, disable e10s-multi for add-ons marked as incompatible. If you have issues with any add-ons, disable e10s completely.
Removed “e10s - (PID)” label from tabs._

grahamperrin:
>> …  If you have issues with any add-ons, …

Refreshment should reduce, not increase, the likelihood of issues.

AjvarXX:
Refreshment will decrease problems as long as addons will be deleted in the process (it will be fixed hopefully).

I personally always turned on e10s even with few unsupported addons.

grahamperrin:
> … addons will be deleted …

That's a different issue …

grahamperrin:
#538

Peacock365:
@grahamperrin 

Does it still occur in Waterfox 56.1 or 56.2 in the same manner? Please verify. If it doesn't happen anymore, you could close the issue.

grahamperrin:
Given the other e10s-related issues, I don't imagine getting a meaningful test result for this issue at this time.

grahamperrin:
## Mac OS X 10.9.5 (Mavericks)

A clean profile '380' created by Waterfox 56.0.2, with addition of one truly incompatible extension: [Mozilla Archive Format](https://addons.mozilla.org/addon/mozilla-archive-format/). 

56.0.2: 

- not bugged.

56.0.3: 

- multiprocess windows reportedly enabled by user
- truly **not** enabled by me
- the extension malfunctions. 

56.1.0

- multiprocess windows reportedly disabled forcibly 
- not forced by me
- the force persists following a refresh. 

56.2.0: 

- multiprocess windows reportedly disabled forcibly 
- not forced by me
- the force persists following a refresh. 

56.2.1: 

- multiprocess windows reportedly disabled forcibly 
- not forced by me
[...]

Issue #392, opened 2018-01-19
Refresh Waterfox: graceful handling of extensions.getAddons.cache.enabled and its cached data


grahamperrin:
In [Refresh Waterfox: the dialogue should not suggest that add-ons will be removed · Issue #390 · MrAlex94/Waterfox](https://github.com/MrAlex94/Waterfox/issues/390) the before-and-after screenshots show: 

- 252 extensions before
- 252 extensions after. 

Unfortunately, it will be necessary for me to reinstall many of those – because the [data cached from AMO, which I spent hours regaining after switching from Firefox to Waterfox](https://github.com/MrAlex94/Waterfox/issues/300#issuecomment-356801957), was: 

- lost by the refresh
- **and not** regained by reusing the old data, which was left on the desktop after the refresh. 

## Enhancement request

Improve the refresh dialogue(s); ask whether information about add-ons should be discarded. 

Alternatively, maybe more simply: 

- if `extensions.getAddons.cache.enabled` was already `true` then (without changing any dialogue) make it `true` for refreshment, and preserve the data that was cached through that preference.

grahamperrin:
https://github.com/MrAlex94/Waterfox/commit/8132766ef286208d7ea5137a1ee49b93bb8d4161#diff-d7a0401e94f68887594ad1d1d9f9c28dR45 noted with thanks.

grahamperrin:
> - if `extensions.getAddons.cache.enabled` was already `true` then (without changing any dialogue) make it `true` for refreshment, and preserve the data that was cached through that preference. 

OK, it's `true` following a refresh 👍 with 56.1.0 on Mac OS X. 

I'll leave this issue open as a reminder for me to raise a pull request for a minor inconsistency with the wording of the Waterfox refresh routine.

grahamperrin:
Tested with 56.2.0, build 20180514223240 on FreeBSD-CURRENT. 

Handling was graceful so I'll close this issue. Sorry for not testing sooner. 

----

Over to UX issue 562 for the on-screen wording, which is not yet consistent with the graceful handling.

Issue #391, opened 2018-01-19
Refresh Waterfox: appearance add-ons (themes) are removed


grahamperrin:
In [Refresh Waterfox: the dialogue should not suggest that add-ons will be removed · Issue #390 · MrAlex94/Waterfox](https://github.com/MrAlex94/Waterfox/issues/390) the before-and-after screenshots show: 

- fourteen appearances before
- three appearances after. 

If I recall correctly: 

- the three that were preserved were the three that were integral to Firefox 56.0.2 (I used Firefox to create the profile).

grahamperrin:
From <https://github.com/MrAlex94/Waterfox/issues/376#issuecomment-357485743>: 

> …  ‘Refresh’ setting utilising the profile import file, which I will rewrite to perform better …

From <https://www.reddit.com/r/waterfox/comments/7ou9na/importing_firefox_profile_was_a_disaster/dschvrk/>: 

> 56.1 will include a completely re-written profile importer. …

grahamperrin:
IIRC this was fixed. Sorry for not closing in good time.

grahamperrin:
Closed in error. A refresh of 56.2.0, build 20180514223240, removed a theme. 

Meta, tracking: #538

MrAlex94:
Themes are meant to be removed.

grahamperrin:
Interesting, thanks. 

Re: #187 I probably have zero experience with complete themes. A wild guess … are some of them so problematic that it's sane to remove all, as a precaution, when refreshing?

Issue #390, opened 2018-01-19
Refresh Waterfox: the dialogue should not suggest that add-ons will be removed


grahamperrin:
With Waterfox 53.0.3_2 on FreeBSD-CURRENT: 

![2018-01-19 07 35 17 refresh waterfox](https://user-images.githubusercontent.com/192271/35139532-72117f12-fceb-11e7-95a7-0942505f9d9c.png)

> ## Refresh Waterfox
> 
> Start fresh to fix problems and restore performance. 
> 
> This will: 
> 
> - remove your add-ons and customisations
> - restore your browser settings to their defaults
> 
> `[[Cancel]]    [Refresh Waterfox]`

## Actual result of refreshment

Extensions, user styles, plug-ins, dictionaries and languages are (loosely speaking) not removed. I love the convenience, the preservation.

An example, before and after refreshment: 

![2018-01-14 16 20 54](https://user-images.githubusercontent.com/192271/35139810-ada4adfa-fcec-11e7-8bfa-ace22bee1a8d.png)

![2018-01-14 16 35 41](https://user-images.githubusercontent.com/192271/35139815-b424c890-fcec-11e7-89b6-3efd92ee0a53.png)

![2018-01-14 16 53 52](https://user-images.githubusercontent.com/192271/35139821-b7eef072-fcec-11e7-95b4-6834680c9804.png)

Side note: the third screenshot – all pages failing to load – is symptomatic of [e10s: multiprocess enabled with incompatible add-ons after refreshing waterfox 56.0.3 · Issue #380 · MrAlex94/Waterfox](https://github.com/MrAlex94/Waterfox/issues/380).

# Issue (enhancement request)

Change the wording in the dialogue, to not suggest that add-ons will be removed.

grahamperrin:
From today's [Waterfox's current status …](https://www.reddit.com/r/waterfox/comments/818z1k/waterfoxs_current_status/): 

> The `Refresh` button has been fixed in `about:support`. You will now be able to reset your profiles while keeping only your important data (same as Firefox).

To clarify: will the fix cause refreshment to remove add-ons from a Waterfox profile (same as Firefox 56.0.2 and Firefox Quantum)?

@MrAlex94 please advise – thanks.

MrAlex94:
No it won’t remove extensions, but it may reset their preferences.

grahamperrin:
Sane. Excellent! Thanks. 

[...]

PR #457, opened 2018-03-03
Update resetProfile.dtd – differs from Firefox

grahamperrin:
Partially address: 

- #390 in particular, "…  it won’t remove extensions, but it may reset their preferences."
- #391
- #392

Further action may include: 

- documentation.

### DIFF
--- a/toolkit/content/resetProfile.xul
+++ b/toolkit/content/resetProfile.xul
@@ -27,9 +27,13 @@
 
   <description value="&refreshProfile.dialog.description1;"></description>
   <label value="&refreshProfile.dialog.description2;"/>
+  <label value="&refreshProfile.dialog.description3;"/>
 
   <vbox id="migratedItems">
     <label class="migratedLabel" value="&refreshProfile.dialog.items.label1;"/>
     <label class="migratedLabel" value="&refreshProfile.dialog.items.label2;"/>
+    <label class="migratedLabel" value="&refreshProfile.dialog.items.label3;"/>
+    <label class="migratedLabel" value="&refreshProfile.dialog.items.label4;"/>
+    <label class="migratedLabel" value="&refreshProfile.dialog.items.label5;"/>
   </vbox>
 </dialog>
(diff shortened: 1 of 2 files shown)


grahamperrin:
Will this expanded `.dtd` require a separate change to the size of the dialogue box(es)? Or does resizing occur automatically? 

More generally: are the bullet points in this `.dtd` a true, reasonably comprehensive reflection of the current state of things? 

I mention "this version of Waterfox" with the assumption that future versions will take different approaches. Losing – [or preserving](https://github.com/MrAlex94/Waterfox/issues/392) – metadata from [a Waterfox-oriented add-ons service](https://github.com/MrAlex94/Waterfox/issues/303); and so on.  

Ideally: I'd like this help routine to be jargon-free. 

More realistically: expect some users to neither know, nor care about, the difference between an _add-on_ and an _extension_ (in the context of this routine) … so a little jargon may be appropriate.

grahamperrin:
## Multi-process defaults

I forgot: 

- [e10s: multiprocess enabled with incompatible add-ons after refreshing waterfox 56.0.3 · Issue #380 …](https://github.com/MrAlex94/Waterfox/issues/380)
- <https://github.com/MrAlex94/Waterfox/issues/397#issuecomment-366545838>.

If the next release (or beta) will not fix #380 then we should: 

1. probably demote _can help_ to (less affirmative) _may help_ in `refreshProfile.dialog.description1`
2. add a bullet point, maybe a final `refreshProfile.dialog.items.label6`, 

> enable multi-process &brandShortName;, even if a extension is incompatible with multi-process. 

– or words to that effect.

MrAlex94:
The changes you've made seem reasonable, but [resetProfile.xul](https://github.com/MrAlex94/Waterfox/blob/master/toolkit/content/resetProfile.xul) needs to reference the new definitions as well. The [CSS](https://github.com/MrAlex94/Waterfox/blob/master/toolkit/content/resetProfile.css) may need adapting as well, if the options are too long.
Not sure what to do about mentioning multi-process though 🤔.

grahamperrin:
At a glance, we might have missed a third line before the indented points: 

2018-03-31 18 03 40 refresh waterfox

I'll aim to fix by making more changes to https://github.com/MrAlex94/Waterfox/pull/488

PR #478, opened 2018-03-20
Expose media.gmp-manager.updateEnabled in about:config by default

laniakea64:
Flipping this pref is required to stop Waterfox from phoning home to Mozilla and Cisco and automatically downloading binaries.  It should be visible in about:config by default.

(To be clear, this patch **does not** change the default value of this pref.)


MrAlex94:
Seems reasonable.

PR #480, opened 2018-03-21
Update some bundled libraries

jbeich:
Tested only on FreeBSD i386/amd64. Cherry-pick if you don't want all.


MrAlex94:
Great job on this. Thanks for the help!

PR #490, opened 2018-03-25
Remove extensions.pocket.enabled and leftover associated code

laniakea64:
Waterfox already removed Pocket itself, this removes the `extensions.pocket.enabled` pref and leftover associated code.


MrAlex94:
Thanks!

PR #503, opened 2018-03-30
Unbreak build on aarch64

jbeich:
#480 regression: missing file. See [error log](http://thunderx1.nyi.freebsd.org/data/latest-per-pkg/waterfox/56.0.4.72_2/111arm64-default.log). Probably affects Android as well.


jbeich:
[Build log](http://thunderx1.nyi.freebsd.org/data/111arm64-quarterly/466007/logs/waterfox-56.1.0_2.log) after the fix. `NEON_FLAGS` is empty on aarch64.

PR #511, opened 2018-04-06
Don't enable Stylo by default

jbeich:
Stylo from Firefox 56 is affected by many Web compatibility and security issues. It also crashes and degrades performance in some situations. Let's disable it at runtime but keep `layout.css.servo.enabled` usable for experienced users.


MrAlex94:
I agree, still needs some work.

jbeich:
A better option is to simply drop `--enable-stylo` by relying on Firefox 56 default:
https://github.com/MrAlex94/Waterfox/blob/67d2b0ee4c226f7d5c45093bed9b3c3832178b08/toolkit/moz.configure#L601-L611

Mikhail-M12:
I have found that layout.css.servo.enabled=true breaks zooming with "Zoom Text Only" enabled.

MrAlex94:
Fixed a conflict and merged. Cheers 👍

decembre:
It seems broke Firebug too:
<a href="https://forum.userstyles.org/discussion/65474/solved-editing-css-with-waterfox-56-01-firebug-new-bug-solution-disable-stylo?new=1">[SOLVED] Editing CSS with Waterfox 56.01+ firebug : new bug ??? [Solution = Disable Stylo]</a>
"_When i edit an userstyle and preview the effect with firebug open , firebug stop to show my CSS code in the htlm panel.
If i reload the page , it show it._"

Issue #496, opened 2018-03-28
about:config privacy settings that should get new defaults in Waterfox.


Peacock365:
Hi there!

I've noticed that some about:config settings in Waterfox need some tweaking. I am deriving (and quoting) my information from the ghacks-user.js release version 56 here:

https://github.com/ghacksuserjs/ghacks-user.js/releases

I'll list the settings + explanation (if needed) here, and the values which I find more appropriate:

**IMPORTANT**

_toolkit.telemetry.unified_ - false

_toolkit.telemetry.cachedClientID_ - empty string

_dom.ipc.plugins.reportCrashURL_ - false;
**disable sending the URL of the website where a plugin crashed**

_dom.ipc.plugins.flash.subprocess.crashreporter.enabled_ - false;
**disable sending Flash crash reports**

_browser.tabs.crashReporting.sendReport_ - false;
**disable sending of crash reports (FF44+)**

_browser.newtab.preload_ - false
_browser.newtabpage.enabled_ - false
_browser.newtabpage.enhanced_ - false
_browser.newtabpage.directory.source_ - empty string;
**disable new tab tile ads & preload & marketing junk**

_network.allow-experiments_ - false;
**disable Mozilla permission to silently opt you into tests**

_experiments.activeExperiment_ - false
**disable all active experiments**

_browser.onboarding.enabled_ - false;
**Onboarding is an interactive tour/setup for new installs/profiles and features. Every time about:home or about:newtab is opened, the onboarding overlay is injected into that page. Onboarding uses Google Analytics, and leaks resource://URIs**

_extensions.formautofill.addresses.enabled_ - false
_extensions.formautofill.available_ - off
_extensions.formautofill.creditCards.enabled_ - false
_extensions.formautofill.heuristics.enabled_ - false;
**disable Form Autofill (FF55+) - Stored data is NOT secure (uses a JSON file)**

Thank you for your attention. This issue was created for discussion & for a potential creation of related commits.

laniakea64:
> _browser.newtabpage.directory.source_ - empty string;

I had to set this one to `data:application/json,{}` to prevent browser console spam.

> _social.enabled_ - false HIDDEN PREF;

Social API is dead, to the point that it was removed from Firefox.  And a quick look at the replacement extension looks Waterfox-compatible.

I'm planning to do a pull request porting bug 1388902 and bug 1406193 to Waterfox.

> _experiments.manifest.uri_ - empty string;

This is already default.

[...]

PR #529, opened 2018-04-18
Make some default settings more privacy-friendly; Fix browser console spam

laniakea64:
Partially address https://github.com/MrAlex94/Waterfox/issues/496


laniakea64:
In light of https://github.com/MrAlex94/Waterfox/issues/496#issuecomment-382464227 , I think some of this should probably be done differently.  Closing.

PR #530, opened 2018-04-18
Make some default settings more privacy-friendly [take 2]

laniakea64:
Fix https://github.com/MrAlex94/Waterfox/issues/496

### DIFF
--- a/browser/app/profile/firefox.js
+++ b/browser/app/profile/firefox.js
@@ -1281,7 +1281,7 @@ pref("browser.newtabpage.rows", 3);
 pref("browser.newtabpage.columns", 5);
 
 // directory tiles download URL
-sticky_pref("browser.newtabpage.directory.source", "");
+sticky_pref("browser.newtabpage.directory.source", "data:application/json,{}");
 
 // activates Activity Stream
 #ifdef NIGHTLY_BUILD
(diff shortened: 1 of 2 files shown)


laniakea64:
Does this code need changing in light of new default for `browser.newtabpage.enhanced`? -

https://github.com/MrAlex94/Waterfox/blob/a52b5eab5aa079a614683ad9442be74a852321b1/browser/modules/DirectoryLinksProvider.jsm#L92-L106

PandaCodex:
I submitted a [patch](https://github.com/MrAlex94/Waterfox/pull/110) while back in reference to [issue](https://github.com/MrAlex94/Waterfox/issues/98) that removed enhanced newtab with mozilla's sponsored tiles even removed the whole DirectoryLinksProvider but was never implemented and was dirtied by commits from alex so needed work too to make viable. so he never used the patch, i could strip it all out again if alex likes but don't want to waste my time if just repeat of last time. patch was originally from Cyberfox

Peacock365:
@PandaCodex 

I am sure Alex would appreciate such a commit (can't speak for him, though), as sponsored tiles don't have _any_ business in _any_ browser.

Some time ago I saw you mentioning the possibility of displaying the add-on version number by default. You said that this could also be ported over from Cyberfox. I think this would be a very welcome improvement, as the current state of affairs (inherited from Mozilla) is rather depressing.

MrAlex94:
At this stage, it's probably worth me going through all the new-tab stuff again.

PR #539, opened 2018-04-25
Update ffmpeg bindings and libwebp

jbeich:
If downstream (e.g., [Arch Linux](https://www.archlinux.org/todo/ffmpeg-40/)) updates **ffmpeg** package to 4.0 it may break H.264/AAC/MP3 playback in Waterfox. Building from source won't help because Gecko bundles headers from various FFmpeg/libav versions.


jbeich:
Oops, forgot to bump date in media/libwebp/MOZCHANGES

MrAlex94:
Great job once again! Thanks.

Issue #405, opened 2018-01-23
some firefox directories still created


Yay295:
56.0.3 clean install, Firefox has never been installed

C:\Users\<user>\AppData\Local\Mozilla\updates\6F940AC27A98DD61
C:\Users\<user>\AppData\LocalLow\Mozilla\Temp-{243ec4a7-ba21-4c80-9a09-4cbfbf2fd1f9}
C:\Users\<user>\AppData\Roaming\Mozilla\Extensions

These directories are empty, but shouldn't be there at all.

grahamperrin:
`~/Library/Caches/Mozilla/updates/Applications/Waterfox/updates`

- from https://pastebin.mozilla.org/9087712
- via https://www.reddit.com/comments/8r66jc/-/e0qa0jo/

Postscript: voting myself down through agreement with a closure elsewhere, https://github.com/MrAlex94/Waterfox/issues/463#issuecomment-398044779

Issue #463, opened 2018-03-10
~/Library/Caches/Mozilla/updates/Applications/Waterfox/ (and other data that's cached on Mac OS X)


grahamperrin:
On Mac OS X, 

- `~/Library/Caches/Mozilla/updates/Applications/Waterfox/`

– is created by Waterfox 56.0.4.1. 

----

I don't imagine any problem arising from use of that path but for tidiness, a future version of Waterfox might use e.g.: 

- `~/Library/Caches/org.waterfoxproject.waterfox/profiles/`
- `~/Library/Caches/org.waterfoxproject.waterfox/updates/`

– and so on. 

## References and side notes

`~/Library/Caches` in Mozilla bug [291033 - Enable support for profile temp directory on local filesystem](https://bugzilla.mozilla.org/show_bug.cgi?id=291033)

File System Programming Guide – File System Basics – [macOS Standard Directories: Where Files Reside](https://developer.apple.com/library/content/documentation/FileManagement/Conceptual/FileSystemProgrammingGuide/FileSystemOverview/FileSystemOverview.html#//apple_ref/doc/uid/TP40010672-CH2-SW6)

Re: sandboxing there's e.g. [Container Directories and File System Access](https://developer.apple.com/library/content/documentation/Security/Conceptual/AppSandboxDesignGuide/AppSandboxInDepth/AppSandboxInDepth.html#//apple_ref/doc/uid/TP40011183-CH3-SW6) … but that's a potentially **massive** side note `:-)`

MrAlex94:
For now, will keep it as is, just so things don't break! But good reference for any huge updates.

Issue #326, opened 2017-12-23
Java plugin: IcedTea-Web activated but not working with Waterfox 56.0.1


grahamperrin:
## Observations

An initial visit to either https://java.com/en/download/installed8.jsp?detect=jre or https://javatester.org/version.html will cause the generic plugin icon to appear in the address bar (to the right of the ⓘ (information) icon). 

Attention to the address bar will allow activation of the plugin. However: 

- the activated plugin is not effective; and 
- **if** allowance is remembered **then** all subsequent loads of the page will not present the icon in the address bar – in other words, no obvious way to end an allowance. 

[2017-12-19 09:02.ogv.zip](https://github.com/MrAlex94/Waterfox/files/1584059/2017-12-19.09.02.ogv.zip)

### If Adobe Flash Player is also installed

`about:addons` options for IcedTeaWeb include an option that is unexpected: 

> Block dangerous and intrusive Flash content ☑️ 

– the same option appears (and is expected) for Shockwave Flash. 

<https://www.ecosia.org/search?q=%22Block+dangerous+and+intrusive+Flash+content%22+%2B%22IcedTea%22> finds nothing. 

## From prior issues

<https://github.com/MrAlex94/Waterfox/issues/313#issuecomment-351575099> (2017-12-14): 

> … does not present its pink rectangle … (**postscript**) neither does it present a plugin icon in the address bar, although I did activate the IcedTea-Web plugin there a day or two ago. At https://java.com/en/download/installed8.jsp the verification routine (for Firefox ESR) did initially present the plugin icon – during and after activation – but subsequent visits to the page are without the icon in the address bar.) … 
 
<https://github.com/MrAlex94/Waterfox/issues/322#issuecomment-353695114> (2017-12-23): 

> …  openjdk and my result is in Waterfox about:plugins , 
> 
> ![icedtea-web](https://user-images.githubusercontent.com/192271/34319219-6a887dbc-e7d4-11e7-925a-71935536d4a3.png)
> 
> looks good, but still don't work 
> 
> ![java-verifying](https://user-images.githubusercontent.com/192271/34319225-7e98b330-e7d4-11e7-805c-d8ea6eb61304.png)

– the address bar is without the icon for the plugin. 

## Environments tested by me

1. Waterfox 56.0.1 on FreeBSD-CURRENT
2. Waterfox 56.0.1 on Kubuntu 17.10.

The FreeBSD environment: 

<details>
 
```text
[...]

PR #547, opened 2018-05-11
Prevent Firefox directories from being used

ghost:
Prevents some leftover directories of Mozilla Firefox from being created and used, as mentioned in #405


MrAlex94:
Did you managed to do any tests to see if it affects anything? And if so, any specific platform?

Peacock365:
We should check whether or not this breaks anything, especially in relation to the Sync feature.

grahamperrin:
Might any of this PR address #326? (Just curious. Very low priority for me.)

grahamperrin:
~Firefox

For reference only, from the closing comment under [~/Library/Caches/Mozilla/updates/Applications/Waterfox/ (and other data that's cached on Mac OS X) · Issue #463 …](https://github.com/MrAlex94/Waterfox/issues/463): 

> For now, will keep it as is, just so things don't break! … 

+1 to not messing with update routines.

Issue #548, opened 2018-05-11
Miscellaneous links that need to be changed in Waterfox.


Peacock365:
Hey there!

This is a short list of links that still point to wrong targets in Waterfox, and which certainly need to be changed:

- [x] In the Sync submenu in the settings, the Android link does not point to Waterfox's Play Store entry, but to a Mozilla page.

- [x] In the Sync submenu in the settings, it still says "Android _and iOS_". Due to Waterfox not having an iOS app, the "_and iOS_" part needs to be removed. Note though that this might have implications for the language packs which possibly also need to be modified.

- [x] In the settings, the "Help" footer at the bottom left of the page points to a Mozilla page instead of https://www.reddit.com/r/waterfox/ (which would be my suggestion).

- [x] Same goes for the "Help" entries under the Help menu  in the main menu bar and in the hamburger menu.

- [x] Under the Help menu in the main menu bar, there is a "Submit Feedback" entry. If you want to keep it, this should probably also point to the Waterfox subreddit. Although this would be just a duplicate of the "Help" entry, which may render "Feedback" obsolete.

That's all. Hope those get fixed.

ghost:
Hi!

I just created a PR at #549, which will fix the first two issues, that you mentioned. 

And I'll take a look at the other issues regarding the help entries. I also think we should let them point to [https://www.reddit.com/r/waterfox/](https://www.reddit.com/r/waterfox/) instead, since it already does that in the android app, I think.

grahamperrin:
A little duplication here; 

> … Under the Help menu in the main menu bar, there is a "Feedback" entry. …

That's [Waterfox ▸ Help ▸ Submit feedback… · Issue #399 …](https://github.com/MrAlex94/Waterfox/issues/399)

> … "Help" entries under the Help menu …

That's [Waterfox ▸ Help ▸ Waterfox help · Issue #400 …](https://github.com/MrAlex94/Waterfox/issues/400)

Peacock365:
@Toromino 

Thanks for having fixed the first ones already. This was rather quick. Amazing!

@grahamperrin 

Thanks for adding context, I didn't know that issues were already created for some of them. When the fixes arrive, you may close them as well.

grahamperrin:
AOK 👍

Peacock365:
@grahamperrin 

#400 is already fixed at large though, you may want to close this one.

Peacock365:
@Toromino 

[...]

Issue #538, opened 2018-04-25
🐞 Waterfox Issue List Tracker


Peacock365:
_Hey there!_

_This is a list of Waterfox bugs yet unresolved. No guarantee for completeness. Should some of them be resolved by now, feel free to inform me. I will delete the corresponding entries in this case. This issue was created in an attempt to create an overview and as a potential starting point for sorting them out._

### Waterfox General
- [ ] Portable updater deleting inappropriately? 
- [ ] Build Waterfox on linux with static musl libc
- [ ] Add README to Linux tarball
- [x] ANGLE -> MoltenVK for macOS
- [ ] Checksum for file downloads
- [ ] Revert Bug 606655
- [ ] libc++ for macOS builds (https://github.com/parrotgeek1/FirefoxLegacy)
- [ ] 1405031 - Support additional click events for browserAction and pageAction
- [ ] https://github.com/MrAlex94/Waterfox/issues/1265
- [ ] https://bugzilla.mozilla.org/show_bug.cgi?id=1373490
- [ ] https://bugzilla.mozilla.org/show_bug.cgi?id=1419264
- [ ] https://bugzilla.mozilla.org/show_bug.cgi?id=1596894
- [ ] https://bugzilla.mozilla.org/show_bug.cgi?id=1419264
- [ ] Re-instate use of [WebCompat](https://github.com/mozilla/gecko-dev/tree/esr60/browser/extensions/webcompat) add-on to help fix website issues, including [UA](https://wiki.mozilla.org/Compatibility/Go_Faster_Addon/Override_Policies_and_Workflows#User_Agent_overrides)!

### Waterfox Classic

- [ ] Waterfox/Mac dictation bug https://github.com/MrAlex94/Waterfox/issues/85
- [ ] On Linux: .config/waterfox | https://github.com/MrAlex94/Waterfox/issues/243 | https://bugzilla.mozilla.org/show_bug.cgi?id=259356
- [ ] Extensions break on path change https://github.com/MrAlex94/Waterfox/issues/329
- [ ] Program description on Windows misleading https://github.com/MrAlex94/Waterfox/issues/387
- [ ] Spellchecking bug https://github.com/MrAlex94/Waterfox/issues/374
- [ ] Autoshortening of hamburger menu fails | https://github.com/MrAlex94/Waterfox/issues/373
| **POSSIBLE SOLUTION**: https://bugzilla.mozilla.org/show_bug.cgi?id=1415921#c0
[...]

PR #573, opened 2018-05-19
"Help" buttons shall lead to Waterfox's reddit page -> #548

ghost:
As proposed in #548 this PR changes the "Help" footer at the bottom of the settings, the "Waterfox Help" button in the hamburger menu and in the menu bar. They'll now lead to Waterfox's reddit page instead of Mozilla's help website.

### DIFF
--- a/browser/base/content/baseMenuOverlay.xul
+++ b/browser/base/content/baseMenuOverlay.xul
@@ -43,7 +43,7 @@
 #endif
       <menupopup id="menu_HelpPopup" onpopupshowing="buildHelpMenu();">
         <menuitem id="menu_openHelp"
-                  oncommand="openHelpLink('firefox-help')"
+                  oncommand="openNewTabWith('https://www.reddit.com/r/waterfox');"
                   onclick="checkForMiddleClick(this, event);"
                   label="&productHelp2.label;"
                   accesskey="&productHelp2.accesskey;"
--- a/browser/components/preferences/in-content-new/preferences.js
+++ b/browser/components/preferences/in-content-new/preferences.js
@@ -85,7 +85,7 @@ function init_all() {
   document.dispatchEvent(initFinished);
 
   let helpButton = document.querySelector(".help-button");
-  let helpUrl = Services.urlFormatter.formatURLPref("app.support.baseURL") + "preferences";
+  let helpUrl = "https://www.reddit.com/r/waterfox/";
   helpButton.setAttribute("href", helpUrl);
 
   // Wait until initialization of all preferences are complete before


criztovyl:
Wouldn't it be better to change the underlying links instead of hard-coding Reddit?

While this won't work directly I would suggest to point the links to a new subdomain support.waterfoxproject.org, which then simply redirects everything to Reddit.

criztovyl:
i.e. change `app.support.baseURL` pref from `https://support.mozilla.org/1/firefox/%VERSION%/%OS%/%LOCALE%/` to `https://support.waterfoxproject.org/1/waterfox/%VERSION%/%OS%/%LOCALE%/` and then using server-side rewrites on `support.waterfoxproject.org/*` to redirect all the request (temporarily?) to `https://reddit.com/r/waterfox`.

criztovyl:
Just my thoughts. :)

ghost:
@criztovyl 

I also thought about this. But changing _app.support.baseURL_ would also break all the other help entries. And some of them still hold useful information like "Keyboard Shortcuts".

criztovyl:
Didn't think about that. ^^
I shortly thought about maybe forking the help infrastructure, but I think that's absolutely not the way to go (neither did I find the help data nor is it feasible to maintain that additional infrastructure).

So, just ignore my comments ^^

PandaCodex:
In Cyberfox `app.support.baseURL`  points to there forums and forum documents. They coded in 
```javascript
pref("app.helpdoc.baseURI", "https://support.mozilla.org/1/Firefox/%VERSION%/%OS%/%LOCALE%/");
``` 
which sends users to Mozilla help documents. Could port these changes over to Waterfox so basic `app.support.baseURL`  points to Reddit where all others point to Mozilla threads for the time being until Alex mirrors there help system or implements own. 

Would this be a PR wanted???

ghost:
@PandaCodex

 I think that'd be a better solution than hard-coding our Reddit URL! :)

grahamperrin:
> … suggest to point the links to a new subdomain support.waterfoxproject.org, which then simply redirects everything to Reddit. …

Meta/tracking: #582

grahamperrin:
Meta, tracking: #538

Maybe expand this PR to cover other areas that should refer to the subreddit? 

### about:support (troubleshooting information)

> … support web site.  …

- refers to https://support.mozilla.org/1/firefox/56.2.14/FreeBSD/en-GB/ (on FreeBSD), redirects to https://support.mozilla.org/en-US/products/firefox?as=u&utm_source=inproduct
- should refer to https://www.reddit.com/r/waterfox/

Issue #399, opened 2018-01-20
Waterfox ▸ Help ▸ Submit feedback…


grahamperrin:
Feedback URLs such as https://input.mozilla.org/feedback/waterfox/56.0.3/ currently redirect to Firefox-specific https://qsurvey.mozilla.com/s3/FirefoxInput/

## On the Waterfox side of things

The _Submit feedback…_ menu item should be associated with a different URL (one that will not redirect to Firefox-specific feedback), maybe https://www.reddit.com/r/waterfox/

- or remove the menu item, until there's something like a SurveyGizmo routine for Waterfox (<https://qsurvey.mozilla.com/> redirects to <https://www.surveygizmo.com/>). 

For what it's worth: I favour temporary removal of the feedback item. Focus instead on **Help ▸ … Help** …

## On the Mozilla side

There should be an issue for the inappropriate redirect …

WagnerGMD:
From my point of view, reddit will be one bad choice for matters related to any code. It's totally buggy (the markdown code syntax) and they doesn't seem to care about (none improvements, neither news in the last 2 months).
At the contrary, it could be an idea for the **Submit feedback**. Another way could be the GitHub itself because (for me it's more readable and) the issue has a features called **tags** (or **label**). It will help a lot in the aim to filter these posts (after in the futur).

grahamperrin:
I'm very often frustrated by the deficiencies of Reddit. I suggested Reddit only because it is (or was, [in 2014](https://redd.it/2ov0dr)) preferred by @MrAlex94.

---- 

For different reasons, GitHub can be equally frustrating. 

Features such as labels and milestones do have the potential to add value. However: if not very carefully planned – and if not applied with consistency – things such as labels can make it **more** difficult to find what's required, especially when a repository becomes relatively busy. 

There are notions such as [Suggestion: Introduce a Discussion tab · Issue #44 · dear-github/dear-github](https://github.com/dear-github/dear-github/issues/44) but I don't foresee GitHub becoming well-suited to discussion.

WagnerGMD:
I understand very well @grahamperrin. We might need to think about that one day (Waterfox futur website).

[...]

Issue #336, opened 2017-12-27
about:addons 'Check to see if your plugins are up to date'


grahamperrin:
The linked page https://www.mozilla.org/plugincheck/ no longer performs any check. I suggest removing the link text from Waterfox: 

> Check to see if your plugins are up to date

## Historical

- https://web.archive.org/web/20170802042307/https://www.mozilla.org/en-GB/plugincheck/ treated Adobe Flash Player as commonplace
- https://web.archive.org/web/20170825070945/https://www.mozilla.org/en-US/plugincheck/ no checking
- https://web.archive.org/web/20171202022401/https://www.mozilla.org/en-GB/plugincheck/ no checking

grahamperrin:
Worked around with [Addon Manager Fix Micro](https://github.com/Ulf3000/addon_manager_fix_micro) but (sorry) I can't tell which parts of the code achieve the workaround.

grahamperrin:
Waterfox 56.2.8 on Windows 7 Enterprise: 

![image](https://user-images.githubusercontent.com/192271/54686216-aa494800-4b10-11e9-9d32-6865d79bfe64.png)

![image](https://user-images.githubusercontent.com/192271/54686290-d664c900-4b10-11e9-8db6-df38dd95a783.png)

Probably fixed by https://github.com/MrAlex94/Waterfox/commit/6d47f085ecc2dc8c59e0ae21adcb008d13e0847b

PR #579, opened 2018-05-23
Waterfox Help -> Submit Feedback (removal)

ghost:
As we don't have any feedback page it was suggested, that we remove the "Submit Feedback" entry of Waterfox's help menu. Pointing it towards Reddit wouldn't make sense, as suggested in #573 the "Waterfox Help" entry would point towards Waterfox's Reddit page anyway.

Related topic: #399 
Also mentioned in #548

### DIFF
--- a/browser/app/profile/firefox.js
+++ b/browser/app/profile/firefox.js
@@ -31,7 +31,7 @@ pref("extensions.logging.enabled", false);
 pref("extensions.strictCompatibility", false);
 
 // Disable compatibility checking due to mish-mash of versions.
-pref("extensions.checkCompatibility.56.1", false);
+pref("extensions.checkCompatibility.56.2", false);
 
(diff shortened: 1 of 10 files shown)


grahamperrin:
Thanks folks. 

Confirmed, the Help menu no longer includes _Submit Feedback_ in Waterfox 56.2.0.31 on FreeBSD-CURRENT. 

----

Whenever I see removals-oriented issues (#336 and #399 for example) I quietly wonder whether something might somehow take the place, in the GUI, of what's to be removed. 

Today it occurred to me that the condensed Help menu might accommodate something that's been bubbling under for months: 

- [FAQ page or pages, maybe in-browser · Issue #603 · MrAlex94/Waterfox](https://github.com/MrAlex94/Waterfox/issues/603)

PR #591, opened 2018-05-27
Stop the address trimming in the address bar (show "http://" and "https://" in all cases)

Peacock365:
Stops the address trimming in the address bar, improves the UI, enhances the information given in the URL bar, shows full URL by default (stops unnecessary URL shortening).

### DIFF
--- a/browser/app/profile/firefox.js
+++ b/browser/app/profile/firefox.js
@@ -342,7 +342,7 @@ pref("browser.urlbar.maxCharsForSearchSuggestions", 20);
 pref("browser.urlbar.suggest.history.onlyTyped",    false);
 
 pref("browser.urlbar.formatting.enabled", true);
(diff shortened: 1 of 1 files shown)


laniakea64:
Just curious, why does this pref need to remain "user set" when user's value matches the default value?

Peacock365:
Oops, you are right. I am a bit busy at the moment, so could you please change _sticky_pref_ to normal _pref_ @laniakea64? That would be great. Please also correct https://github.com/MrAlex94/Waterfox/commit/147fc6c9412678f78ca1e838594bd40601e3760c I have created a new pull request with this change and some others: https://github.com/MrAlex94/Waterfox/pull/593

laniakea64:
@Peacock365 Sorry for the delay in responding.  Unfortunately I too am very busy atm and I don't currently have time to do pull requests. :(

PR #624, opened 2018-06-15
Improve SSSE3 scaler.

Peacock365:


MrAlex94:
Is there a Bug ID to accompany this? In fact, most of the patches are lacking one. Also, could I recommend learning [git cherry-pick](https://git-scm.com/docs/git-cherry-pick), it's useful for keeping track of commits, as well makes future merges much easier to handle.

Peacock365:
@MrAlex94 

Most of the commits in fact have no Bug ID, as I ported them from UXP. Some performance improvements implemented there. However, oftentimes they need to be adapted to the changed Gecko 56 code, and may not mirror the original commit in all cases.

In case of the SSE3 scaler, it's luckily 1:1 same: https://github.com/MoonchildProductions/UXP/commit/cd4da92df48738574c832d9badefd9c542fc6149

MrAlex94:
Cool, I’ll try and review them all soon :-)

MrAlex94:
Thanks for pushing this patch, but I've added it to have proper reference. Turns out it's a security bug, but there was no context from the UXP commit.

https://github.com/MrAlex94/Waterfox/commit/83355f6722a4388e8f3fc706c145a2707fe6456c
https://github.com/MrAlex94/Waterfox/commit/c160fa3d0268ba82cb0b25dd42e43340190ef184

PR #628, opened 2018-06-15
Add a check to prevent a crash when CSP directives are used with chro… 

Peacock365:
…me content

Chrome content uses System Principle and this kind of speculative preloading of CSP just won't work.


jbeich:
Rationale is often more important than a patch itself. What's the origin here? If you're the author, please, provide the stacktrace and test case.

grahamperrin:
https://github.com/MoonchildProductions/UXP/commit/7d9e2a108478d1fc79aa25b2ac56b62249e31dc5 was _Add a check to prevent a crash when CSP directives are used with chrome content_,

> Chrome content uses System Principle and this kind of speculative preloading of CSP just won't work.

Issue #553, opened 2018-05-12
Skia has just been updated to match release branch. Please build and test!


MrAlex94:
Hi everyone,

I've [tested out](https://stackoverflow.com/a/35914396/1154632) a new way to keep Waterfox up to date with various projects. 

This is quite a large change, and relates to anything canvas (including WebGL). Please test :-). Currently gfx/skia now matches the release branch of the Mozilla platform. I've done on my platform (macOS 10.13) and seems to work without issue.

Cheers!

grahamperrin:
Thanks! 

Two questions, maybe both are stupid but here goes …

----

The two preferences `gfx.canvas.azure.backends` and `gfx.content.azure.backends` at about:config?filter=azure.backends – if a problem _is_ found with either canvas or content, then would it be sane to experiment with `cairo` in lieu of `skia` for that one preference?

(I never heard of [Skia](https://skia.org/) before today … reading this issue alongside old (2016) <https://forums.freebsd.org/threads/firefox-no-gpu-acceleration-since-50-0.59089/#post-338314>.)

----

If there's a choice of rendering back-ends (e.g. XRender | OpenGL 2.0 | OpenGL 3.1) for compositing, then should the updated Skia be tested against all back-ends?

laniakea64:
So far so good in self build from https://github.com/MrAlex94/Waterfox/commit/941b28140102e33f968ad710548333aee9d1b618 running on Xubuntu 18.04 64-bit.  I play HTML5 games and haven't noticed any problems so far.

Peacock365:
It‘s a great way to keep Waterfox up to date. No problems so far.

I must ask, however, is fixing all the issues mentioned in #538 still on your radar, @MrAlex94 ? I have gone to great lengths to find (hopefully) appropriate fixes for most issues...

grahamperrin:
@Peacock365 if you must bump an issue, please do it in the issue (not in other issues). Thanks.

jbeich:
Builds fine on FreeBSD [10.4 i386](https://ptpb.pw/lO51)/[11.1 amd64](https://ptpb.pw/TuYx).

> new way to keep Waterfox up to date with various projects.

Curiously, it exploded commit history.

```
$ git describe --tags edeb7e6627c40~1 --match 56.1.0
56.1.0-89-g2bb1a86e5dbd6

$ git describe --tags edeb7e6627c40 --match 56.1.0
56.1.0-111168-gedeb7e6627c40
```

[...]

PR #629, opened 2018-06-15
Enable xrender by default.

Peacock365:
Resolves perf regressions over remote X.

### DIFF
--- a/modules/libpref/init/all.js
+++ b/modules/libpref/init/all.js
@@ -4883,7 +4883,7 @@ pref("gfx.apitrace.enabled",false);
 
 #ifdef MOZ_X11
 #ifdef MOZ_WIDGET_GTK
-pref("gfx.xrender.enabled",false);
+pref("gfx.xrender.enabled", true);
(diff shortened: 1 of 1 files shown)


grahamperrin:
From <https://github.com/MrAlex94/Waterfox/issues/553#issuecomment-388657333>: 

>> If there's a choice of rendering back-ends (e.g. XRender | OpenGL 2.0 | OpenGL 3.1) for compositing, then should the updated Skia be tested against all back-ends?

> - XRender is **probably** irrelevant: disabled due to unpredictable performance and interference with OpenGL
> - Skia takes advantage of OpenGL compositing via `gfx.canvas.azure.accelerated=true` (aka SkiaGPU)

I see `gfx.xrender.enabled` `false` by default with e.g. www/firefox 60.0.2,1

@jbeich please, do you know, is the interference specific to FreeBSD?

jbeich:
@grahamperrin, OpenGL compositing is `layers.acceleration.force-enabled=true`. Try benchmarking e.g., via HTML5 games.

PR #644, opened 2018-06-16
Enable blocking of top-level data: navigation.

Peacock365:


laniakea64:
This will break a lot of local testing.  In many cases, it's much simpler to quickly type a data: URI in the address bar, than to spin up a local server, make a webpage, upload it to the local server, and point the browser to the URL.

So what's the advantage?

PR #662, opened 2018-06-17
Remove referrers when opening a link in a new private window.

Peacock365:


grahamperrin:
Contexts: 

- [Open in Private Window](https://addons.mozilla.org/addon/openinprivatewindow/) (2013-05-04) "… By default, Firefox send original referrer when you open a URL in Private Window. With this extension, you can disable it. …"
- Mozilla bug [1409226 - Referer is sent when opening a new private window](https://bugzilla.mozilla.org/show_bug.cgi?id=1409226) RESOLVED FIXED in Firefox 58

PR #669, opened 2018-06-18
Pre-process extensions.js

Peacock365:


MrAlex94:
Why? There are no preprocessor directives in https://github.com/MrAlex94/Waterfox/blob/master/toolkit/mozapps/extensions/content/extensions.js

PR #670, opened 2018-06-18
Never shrink line heights as a result of minimumFontSize.

Peacock365:


MrAlex94:
Are you just porting a bunch of patches from UXP? What’s the context for this patch? Where was this issue first reported?

Peacock365:
The issue was first reported here: https://github.com/MoonchildProductions/Pale-Moon/issues/1614

Well, I am not porting *all* patches from UXP, as a huge part of them are just dealing with the Pale Moon application code (which has nothing to do with the Waterfox application code). Those are just bugfixes and fixes for known performance regressions.

Peacock365:
And I forgot to mention that most patches belonging to UXP are already available in Firefox / Waterfox 56. The patches I ported over are basically the only ones not coming straight from Mozilla and not dealing with the application code.

PR #677, opened 2018-06-19
Add extra check for path traversal sanity v2.

Peacock365:


MrAlex94:
For bugs like this, is it possible to know where they are from/what issue they fix? I've tried searching for this one, but can't find anything direct? The other Mozilla bugs are obviously fairly easy to find and I've verified/merged them, but these ones are a bit ambiguous. I know the other bug you said it was directly from UXP or similar, but are all the ones unmarked from that project?

grahamperrin:
https://github.com/MoonchildProductions/UXP/commit/dda392cd4edb3258889188af5a5644eb8d36aeb7 was _Add extra check for path traversal sanity v2._

PR #689, opened 2018-07-03
Fix identity.mobilepromo.android URL

notpushkin:
The link in `about:preferences#sync` is broken right now as a utm_campaign value is appended to it (“consumers are expected to append a value for utm_campaign”). I've added the `utm_` parameters back, just like in the iOS url. Not sure if this would be helpful in the analytics sense, but at the very least the link works again :)


MrAlex94:
Thanks! I don't have any analytics for it, so doesn't matter that much what the values are.

PR #698, opened 2018-07-13
Change sponsorship logo and link

elof:
Hi Waterfox folk, I'm a developer evangelist over at StackPath. We bought MaxCDN a couple years ago and are currently making steps towards consolidating brands. You don't need to make any changes to continue receiving the support we have been and will continue to give you, but we would appreciate if you change the logo and link to reference StackPath when you talk about the sponsorship. If you could also change it on your site that would be greatly appreciated. Let me know if you have any questions or concerns at all.


MrAlex94:
Sure, no problem! Will change it over ASAP

PR #699, opened 2018-07-14
Bug 1314968

hawkeye116477:
`https://bugzilla.mozilla.org/show_bug.cgi?id=1314968`
One other commit with another fixes should be before for this, obviously someone lost it in hurry :smile:.


MrAlex94:
Oops, this got lost in a barrage of emails! Thanks :-)

PR #778, opened 2018-11-01
[QOL] Disable the black screen transition when going full-screen

misyltoad:
This transition is way too long, and is incredibly irritating.

Me and everyone I know who uses Water/Firefox disables this too.
It would be a great QoL change to get this merged in for default.

Cheers!
 - Josh

### DIFF
--- a/modules/libpref/init/all.js
+++ b/modules/libpref/init/all.js
@@ -5010,8 +5010,8 @@ pref("full-screen-api.allow-trusted-requests-only", true);
 pref("full-screen-api.pointer-lock.enabled", true);
 // transition duration of fade-to-black and fade-from-black, unit: ms
 #ifndef MOZ_WIDGET_GTK
-pref("full-screen-api.transition-duration.enter", "200 200");
-pref("full-screen-api.transition-duration.leave", "200 200");
+pref("full-screen-api.transition-duration.enter", "0 0");
+pref("full-screen-api.transition-duration.leave", "0 0");
 #else
 pref("full-screen-api.transition-duration.enter", "0 0");
 pref("full-screen-api.transition-duration.leave", "0 0");


grahamperrin:
I never noticed a black screen. 

about:config?filter=full-screen-api.transition-duration finds defaults of 0 (zero) with Waterfox 56.2.3 on FreeBSD. I upped the values, and like the effect.

misyltoad:
On FreeBSD I am assuming it compiles with MOZ_WIDGET_GTK or something which is why the default would be zero.

You are welcome to like the effect, however:

It still lags greatly and takes 1.61 seconds (I timed it, and it's way longer than set) to transition for me when set to the default values. Other people I know also experience this same issue. I have tried this across Windows and Linux and the issue still persists.

That's 1.61 seconds that I can't do anything with my computer, or consume the content I asked for!

It adds very little to the browsing experience -- in fact it hinders it and actively prohibits me and others from accessing the content that we chose to make full-screen for a fairly extended period of time.

If something doesn't work properly (eg. the timings don't work correctly due to lag) and it hinders a user, it really shouldn't be enabled by default.

misyltoad:
Not only that, but it makes the video I am watching lag to a crawl while it does the fade in/out animation, it's really painful...

WagnerGMD (changes requested):
**You have forgot this one  !**
```
pref("full-screen-api.warning.timeout",0);
// According to my memories, I believe the original value is :
pref("full-screen-api.warning.timeout",3000);
```

misyltoad:
This doesn't have an effect on the transition, but just the warning that shows up. That doesn't really bother me that much, but yeah it's not really necessary either.

misyltoad:
Helloooooooooooooooooooooooooooooooooooooooooooooooo?

grahamperrin:
Sought other reports of annoyance in the subreddit for Waterfox, I found this from 2016: https://redd.it/415ffm

Back to 2015, https://redd.it/3rzdyg (liking the fading, if not too long) referred to this Mozilla bug (RESOLVED FIXED in Firefox 42): 

- [1185775 - The transition for entering and exiting fullscreen is too long](https://bugzilla.mozilla.org/show_bug.cgi?id=1185775)

More recent: 

- [1209829 - [meta] Tracking bug for Fullscreen API issues](https://bugzilla.mozilla.org/show_bug.cgi?id=1209829)

– and in its [dependency tree](https://bugzilla.mozilla.org/showdependencytree.cgi?id=1209829&hide_resolved=1): 

- [1187769 - [meta] Full screen transition is black for too long](https://bugzilla.mozilla.org/show_bug.cgi?id=1187769)

misyltoad:
Hellooooooooooooooooooooooooo?

regs01:
I find this smooth transition pretty good. While crude Chrome and Edge transitions are quite annoying for eyes. If you have a problem with this due to specific rare OS, well just set settings for yourself. No need for a step back for everyone just because of you.

PR #781, opened 2018-11-02
Switch to twemoji fonts

ghost:
emojione emoji fonts are obsolete, replaced by twemoji fonts like in firefox


hook321:
I think we should take into consideration the fact that the pistol emoji is incorrect in twemoji.

ghost:
sarcasm?

this is not my own initiative, i'm just copying mozilla, i'm assuming the new file is really better.

emojione was made by some company. The company changed the license, so no new free releases will ever be made. This is why the switch happened.

hook321:
I know Mozilla switched emoji fonts, but that doesn't make being non-compliant with Unicode standards okay.

ghost:
no one develops emojione any more.

twemoji have twitter behind them that still develops them. And mozilla builds the file... I think mozilla knows what it's doing....

you talk about this emoji? 🔫
In waterfox and firefox i see a normal gun.
I might mixed up the font files when i was testing though....
various companies like google and apple use a watergun for PC reasons.... This is what you mean by incorrect?

hook321:
In version 2.6 they switched it to a watergun (https://emojipedia.org/twitter/twemoji-2.6/changed/). It's incorrect because the emoji is named pistol, not watergun. It has nothing to do with politics for me, if someone had just proposed to the Unicode consortium to change the emoji to a watergun and the proposal was accepted then I would have been fine with this. There's a right way and a wrong way for people to change the meaning of emojis, and ignoring standards isn't the right way.

ghost:
They all use a stupid watergun now. 
https://blog.emojipedia.org/all-major-vendors-commit-to-gun-redesign/
Is that standard enough for you?

hook321:
It's not up to them to decide on standards, it's up to the Unicode Consortium. We could potentially fork twemoji and keep all changes except for the change to the pistol emoji.

ghost:
it's very nice what the standard say.

In practice Apple, Google, Microsoft, Facebook, Twitter, Samsung all have the watergun. With out counting all those that just copy pasted their fonts. Ubuntu has the google fonts, while, Firefox has the twitter fonts.....

Issue #563, opened 2018-05-16
About the bookmarks and the interface of any websites 


WagnerGMD:
Hello,

I'm just wondering if you could :
- create inside the profil one folder called **chrome** ?
- and add these files : **userContent.css** and **specially the userChrome.css** ?

To resume, sometimes the people just want the simple solution. It might be a better idea than to have to install an addon.

What about **one button to export or import the bookmarks ?** I think about since several years but Mozilla seem's far away to listen.
By the way, could you explain how to get the right extension ? Because I want a **.jsonlz4** instead of (the usual or normal) **.json** file. Just open the profil (inside the folder **bookmarkbackups**) and you will see the difference (the size is very light).

Kind regards.

**PS :** I didn't follow the story (sound bad) about [Stylish](https://addons.mozilla.org/en-US/firefox/addon/stylish/). But apparently there is (at least) one fork (Stylus ? etc).
Otherwise there is also [the CustomCSSforFx](https://github.com/Aris-t2/CustomCSSforFx/releases/) as another solution but that's too much.

criztovyl:
I am currently using this approach with Waterfox to make it play nice with GTK dark theme enabled.

WagnerGMD:
**As reminder :**

1. At first, you must create one folder called **chrome** (inside the **profile**).
2. Then create the file : **userContent.css**.
3. You need one exemple ? Just recopy the code (as below).
4. Don't forget to restart the browser (Waterfox, Firefox, Cyberfox, etc).
```
@namespace xul url(http://www.mozilla.org/keymaster/gatekeeper/there.is.only.xul);
@namespace html url(http://www.w3.org/1999/xhtml);
@-moz-document domain("reddit.com")
{
	.fggbYQ,.fQMRlP,#hamburgers,.bGDHLv,.cxXdcf{display: none !important;}
	#eu-cookie-policy,#uapp-banner{display: none !important;}
	.fOVPWD,.VWHvb{margin-left: 0 !important;}
}
@-moz-document domain("stackoverflow.com")
{
	#sidebar,.z-banner,.vote,#dfp-tlb,#dfp-mlb,.old-hero._fixed{display: none !important;}
	.post-layout{display: block !important;width: 1100px !important;}
}
@-moz-document domain("paypal.com")
{
	#gdprCookieBanner{display: none !important;}
}
@-moz-document domain("github.com")
{
	.container{width: 1200px !important;}
	.discussion-timeline{width: 950px !important;}
}
[...]

PR #784, opened 2018-11-06
general.useragent.override.github.com

grahamperrin:
To suppress the banner that now appears for Waterfox 56.2.5 (treated as an old unsupported version of Firefox): 

2018-11-05 23 22 10 issues mralex94-waterfox

`Mozilla/5.0 (rv:60.0) Gecko/20100101 Firefox/60.0`

- platform agnostic
- 60 (e.g. ESR) is enough to suppress the banner.

https://help.github.com/articles/supported-browsers/

### DIFF
--- a/modules/libpref/init/all.js
+++ b/modules/libpref/init/all.js
@@ -28,6 +28,7 @@ pref("general.useragent.compatMode.firefox", false);
 // This pref exists only for testing purposes. In order to disable all
 // overrides by default, don't initialize UserAgentOverrides.jsm.
 pref("general.useragent.site_specific_overrides", true);
+pref("general.useragent.override.github.com", "Mozilla/5.0 (rv:60.0) Gecko/20100101 Firefox/60.0");
 #ifdef XP_WIN
 pref("general.useragent.override.addons.mozilla.org", "Mozilla/5.0 (Windows NT 10.0; Win64; rv:57.0) Gecko/20100101 Firefox/57.0");
 #endif


grahamperrin:
Cross reference: 

- https://www.reddit.com/comments/9drjqx/-/e94m0yv/ 

– with confirmation, from another user, that `60` works (where e.g. `63` failed).

grahamperrin:
Re: https://www.reddit.com/comments/9unobl/-/e95nctl/ an inferior, platform-specific string is also effective: 

- `Mozilla/5.0 (X11; Linux x86_64; rv:56.0) Gecko/20100101 Firefox/56.0`

– but the condensed version is not: 

- `Mozilla/5.0 (rv:56.0) Gecko/20100101 Firefox/56.0`

Gut feeling, without knowing which versions might be outdated next by GitHub: go for `60`.

WagnerGMD:
In the meanwhile, as you can assume there is [another solution](https://github.com/MrAlex94/Waterfox/issues/563#issuecomment-395307955) (just below) : 
```
@-moz-document domain("github.com")
{
	.unsupported-browser{display:none !important;}
}
```
From my point of view, it's probably the easier way.

chylex:
@WagnerGMD GitHub refuses to load javascript if the user agent has an "unsupported" version, which breaks many parts of the website. Hiding the banner doesn't solve that.

grahamperrin:
Also, a `userContent.css` workaround is not a good fit with a pull request

ChiaYen-Kan:
i try to use 

"Mozilla/5.0 (Windows NT 10.0; Win64; rv:60.0) Gecko/20100101 Firefox/60.0" 

it can't work

some function is also can't be used
1. "Switch branches/tags" can't show any branch/tag
2. in edit area Write/Preview can't switch

johny-mnemonic:
@kchiayen Why trying something different? I can confirm, that what is in this PR works fine. 
What you describe means, that Github treats your browser as unsupported => it didn't recognize your useragent string as supported browser.

ChiaYen-Kan:
> 
> 
> @kchiayen Why trying something different? I can confirm, that what is in this PR works fine.
> What you describe means, that Github treats your browser as unsupported => it didn't recognize your useragent string as supported browser.

I know what are difference, my preference item name is wrong, so github can't see custom useragent
when set to "general.useragent.override.github.com" is work fine
thank you

ghost:
Oh god, more user agent madness. Just decide to regularly copy paste the latest firefox useragent already....

grahamperrin:
Not madness.

Use of GitHub is promoted at the home page and elsewhere. A pull request is appropriate.

`60`. 

IIRC (away from computers for a few days) a `64` string did not work for at least one user.

ghost:
64 is currently beta...

the user agent string changed multiple times already for ad hoc reasons. You just add an other ad hoc reason, tomorrow it will be yet something else.....

Simply using the current Firefox string will not work?

chylex:
When I set the version number too high, I started getting JS errors about features that only started being supported in those versions, so it looks like GH is checking UA instead of feature availability.

On top of that, some UAs break for no apparent reason, for example:  
`Mozilla/5.0 (X11; Linux x86_64; rv:56.0) Gecko/20100101 Firefox/56.0` works fine for me, but  
`Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:56.0) Gecko/20100101 Firefox/56.0` doesn't.

hook321:
I think we should at least attempt to ask @github to whitelist Waterfox's useragent. They may say no, but we should at least ask.

johny-mnemonic:
@hook54321 there already was such [attempt](https://github.com/browser-update/browser-update/issues/397), but it was before Waterfox started using its own useragent string.
So I guess it's time to try it again.

grahamperrin:
@johny-mnemonic the Browser-Update.org service: 

- is unrelated to GitHub
- [does whitelist Waterfox](https://github.com/browser-update/browser-update/issues/416#issuecomment-434983087).

grahamperrin:
> … we should at least attempt to ask @github to whitelist Waterfox's useragent. …

+1
although I shouldn't expect a response here (I imagine that the `@github` ID is overused for contact purposes). 

I made made a request using the contact form.

grahamperrin:
@kchiayen wrote: 

> "Mozilla/5.0 (Windows NT 10.0; Win64; rv:60.0) Gecko/20100101 Firefox/60.0"

> …
> 2. in edit area Write/Preview can't switch

I did not have that problem with 56.2.5 on FreeBSD with the non-BSD string, 

`Mozilla/5.0 (Windows NT 10.0; Win64; rv:60.0) Gecko/20100101 Firefox/60.0`

Whatever value you choose for `general.useragent.override.github.com`, take care to **not** surround it with `"…"` quotation marks. 

(Quotation marks are proper in the context of https://github.com/MrAlex94/Waterfox/pull/784/files#diff-61cac4855c1bb1dd3e264ed799753d4fR31 but the marks must **not** be used when setting the value manually.)

grahamperrin:
> … some UAs break … for example: … `Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:56.0) Gecko/20100101 Firefox/56.0` …

Confirmed (with 56.2.5 on FreeBSD-CURRENT), for example I can not preview this comment whilst drafting. 

The same limitation with `Mozilla/5.0 (X11; FreeBSD amd64; rv:56.0; Waterfox) Gecko/20100101 Firefox/56.2.5`. 

----

There's probably a good reason for GitHub's exceptional treatment of Linux-specific `Mozilla/5.0 (X11; Linux x86_64; rv:56.0) Gecko/20100101 Firefox/56.0` …

ACyphus:
Hey folks 👋 

My name is Alex and I work on the Technical Support team @github.

I'm following up on the ticket that @grahamperrin opened with us. 

GitHub is now whitelisting the Waterfox browser based on the contents of the UA. Waterfox will need to contain `Waterfox` in the UA for this whitelisting to apply. 

GitHub does not usually serve JavaScript to unsupported browsers because that can cause things to be _more_ broken than gracefully falling back to a no-JS state.

I hope this helps.

chylex:
> GitHub does not usually serve JavaScript to unsupported browsers because that can cause things to be more broken than gracefully falling back to a no-JS state.

I'm really not a fan of this decision, especially since I've run into issues with GitHub doing something unwanted without JS where it'd be better if it didn't do anything at all - for example, I can write a draft for a release completely fine without JS, but trying to save it as a draft *actually publishes the release and runs related webhooks*, which is much worse than if the Save Draft button didn't do anything.

ACyphus:
> I'm really not a fan of this decision, especially since I've run into issues with GitHub doing something unwanted without JS where it'd be better if it didn't do anything at all - for example, I can write a draft for a release completely fine without JS, but trying to save it as a draft _actually publishes the release and runs related webhooks_, which is much worse than if the Save Draft button didn't do anything.

Thanks for the feedback @chylex. The attempt to save a draft release actually publishing the release sounds like a bug. Would you mind emailing support@github.com with steps to reproduce and a few screenshots? We can then work on getting that resolved.

grahamperrin:
@ACyphus thank you! 

@MrAlex94 would you like to close this PR, without merging?

ghost:
@grahamperrin You can close it, same with issues. If you started them, you'll see a close button next to the comment button.

grahamperrin:
@pirate486743186 yeah, I left this open only in case Alex wants to do anything else with the PR. 

(I don't imagine any adapted use of the PR, but I don't have the big picture.)

grahamperrin:
Without the override, there _is_ the potential to shoot oneself in the foot however to the best of my knowledge, this has happened only once. 

There's an associated rant https://www.reddit.com/r/waterfox/comments/ahkvqa/-/eefrmkp/?context=1 however that's about GitHub, not Waterfox, so I'll close this PR.

PR #812, opened 2018-12-05
Fix the "CraSSh" pure-CSS DoS vulnerability in default configuration

laniakea64:
See https://github.com/MoonchildProductions/UXP/issues/891 and https://cras.sh/ .  I tried the PoC in Waterfox 56.2.5 in a VM, and it totally froze the entire VM!

This pull request is the patch from UXP.  It fixes the issue in Waterfox if `layout.css.servo.enabled` is left at its default value of `false`.  Looks like a separate fix will be needed for the case when `layout.css.servo.enabled` is set to `true`.

### DIFF
--- a/layout/style/nsCSSParser.cpp
+++ b/layout/style/nsCSSParser.cpp
@@ -1412,6 +1412,9 @@ protected:
 
   // All data from successfully parsed properties are placed into |mData|.
   nsCSSExpandedDataBlock mData;
+  
+  // Value to make sure our resolved variable results stay within sane limits.
+  const int32_t MAX_CSS_VAR_LENGTH = 10240;
 
 public:
   // Used from nsCSSParser constructors and destructors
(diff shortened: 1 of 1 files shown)


hook321:
In Firefox 63.0.1 it just crashes the tab for me. I'm not sure if that counts as fixed or not though.

laniakea64:
@hook54321 That's not fixed behavior.  If it's fixed, the PoC page should say "If you can see this, your browser doesn't support modern CSS or the developers have fixed CSS variable handling".

MrAlex94:
Thanks will test and merge

laniakea64:
> Looks like a separate fix will be needed for the case when `layout.css.servo.enabled` is set to `true`.

I think the Mozilla bug for this is https://bugzilla.mozilla.org/show_bug.cgi?id=1510862

Unfortunately the patch doesn't apply as-is to Waterfox, and I don't know Rust so have no idea how to port it to Waterfox.

laniakea64:
UXP followed up their CraSSh patch with https://github.com/MoonchildProductions/UXP/commit/dfce7fa7b664815bbce363d252033094a0dfe843 .  Should Waterfox have this too?

PR #843, opened 2019-01-18
Add Sortable Folder and Path Columns to the Bookmarks Search

bilgincoskun:
This pull request aims to solve https://bugzilla.mozilla.org/show_bug.cgi?id=196509 and https://bugzilla.mozilla.org/show_bug.cgi?id=469421 issue. There was an extension for this in the past but since sorting of columns is done on lower level than the extension folder columns were unsortable. 
This extension adds two columns (name of the parent folder and full path of the parent folder) to the bookmarks search which both are sortable. These columns appear empty when there is no search since all shown links are from the same folder. 
I tested these patches in my daily usage with pretty large bookmarks (180k+) and did not see any noticable performance problem or error.

### DIFF
--- a/browser/components/places/content/places.xul
+++ b/browser/components/places/content/places.xul
@@ -396,6 +396,13 @@
             <splitter class="tree-splitter"/>
             <treecol label="&col.lastmodified.label;" id="placesContentLastModified" anonid="lastModified" flex="1" hidden="true"
                       persist="width hidden ordinal sortActive sortDirection"/>
+            <splitter class="tree-splitter"/>
+            <treecol label="&col.parentfolder.label;" id="placesContentParentFolder" anonid="parentFolder" flex="1" hidden="true"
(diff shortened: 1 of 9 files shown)


MrAlex94:
Great work, thanks!

rjt:
180,000 bookmarks and no noticeable performance problem, awesome!  

Will this allow one to search for "Parent Folders" by name?
Show all "ParentFolders" that a duplicated URLs reside in?
Jump to the "Parent Folder"?

bilgincoskun:
> Will this allow one to search for "Parent Folders" by name?

No.
> Show all "ParentFolders" that a duplicated URLs reside in?

It does not deduplicate search results, so If there are duplicate links, it shows all duplicates with the folders they are in.

> Jump to the "Parent Folder"?

No,but there is Go to Parent Folder extension in the Classic Add-ons Archive for that.

Issue #235, opened 2017-10-05
About the leak (resource)


WagnerGMD:
Hello,

I'm wonder when do you plan to fix this trouble ? If you want to check it, you just need to open [this page ](https://browserleaks.com/firefox) and you will get the same result as on [this picture](http://nsa39.casimages.com/img/2017/10/05/171005030153522808.png).

In the past, that's right the addon [No Resource URI Leak_v1.1.0](https://addons.mozilla.org/en-US/firefox/addon/no-resource-uri-leak/) was a good solution. But it doesn't work anymore (since Firefox_v55 or Firefox_v56). Despite the fact : yes one [new version](https://raw.githubusercontent.com/earthlng/testpages/master/no_resource_uri_leak-1.1.1-an%2Bfx%2Bsm%2Btb.xpi) exist, it wouldn't be better to just include the patch ? Because it was already done with Firefox_v57 and I don't see the point to wait (or avoid).

Kind regards.

jugi1:
I agree with WagnerGMD.
Any solution or alternative?

MrAlex94:
It's patched in v56.

jugi1:
Working now.
Thanks.

L-a-n-g-o-l-i-e-r-s:
@MrAlex94 
It's not fully patched, the test page shows me in comparison to No Resource URI Leak 1.2.2 (Custom version from a repository topic somewhere on the ghacks js file github.) 

While the extension is disabled, it isn't revealing the extensions any more in WF 56, however it is revealing the following without the extension enabled: Platform Detection, Default Locale, Firefox ESR Channel, firefox.js,  firefox-branding.js,  firefox-l10n.js,  webide-prefs.js and greprefs.js, (The js files have hashes which are uniquely identifiable, it is able to not only count the number of strings but display all of the settings. For example, firefox-branding.js contains links to waterfox that identify the browser as not being MFF) all of these are not exposed when the extension is enabled.

I don't know if it's 100% fixed in FF 57, I have heard some mention this, but I figured I would mention this. 

https://browserleaks.com/firefox 🎱 

--

WagnerGMD:
Perhaps there is something wrong in **No Resource URI Leak_v1.2.2** ? I haven't heard about (but on the moment it was fine with **No Resource URI Leak_v1.1.1**)
It's true about Firefox_v57, because a few weeks ago I haven't met any trouble. So as today I will assume nothing has change.

[...]

Issue #178, opened 2017-08-07
User-Agent: Waterfox


iffin2l:
I'd like to know if it's possible to remove Waterfox after Firefox agent to avoid identification
Most people don't use Waterfox. I believe it's safer to leave as if it was default Mozilla Firefox

WagnerGMD:
According to this [reddit topic](https://www.reddit.com/r/waterfox/comments/6of9ct/bug_unable_to_access_website_since_upgrade_to/), there is also another trouble (related to the user-agent).
Because (I had check again today and) it's seem to be true, CloudFlare block the browser.

iffin2l:
I believe it would be both safer and better to change from _'Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:54.0) Gecko/20100101 Firefox/54.0.0.1 Waterfox/54.0.0.1'_ to _Mozilla/5.0 (Windows NT 6.1; Win64; x64; rv:55.0) Gecko/20100101 Firefox/55.0(Firefox/**current release**)_

MrAlex94:
@iffin2l, do you mean Mozilla/5.0 (Windows NT 6.1; Win64; x64; rv:55.0) Gecko/20100101 Firefox/55.0 _(Waterfox/current release)_ ?

iffin2l:
@MrAlex94 Hi Alex, I thought it could be a good thing to let as if it is a default Firefox. Like this _Mozilla/5.0 (Windows NT 6.1; Win64; x64; rv:55.0) Gecko/20100101 Firefox/55.0_
Or.. perhaps having two options, the one with the agent _Mozilla/5.0 (Windows NT 6.1; Win64; x64; rv:55.0) Gecko/20100101 Firefox/55.0 Waterfox/55.0_ and the other one without the Watefox(current release). This is strictly my privacy concern only. Nothing that would break any website I believe
Thank you

WagnerGMD:
Just to inform. Right now, I'm able to access to [pricehipster.com](https://pricehipster.com/). So I will suppose the trouble with **CloudFlare** might be gone.

juneyourtech:
The User Agent Switcher add-on solves many of the issues I have, though I don't use Waterfox.

Firefox for Android has an about:config option to change user agents, like this:

general.useragent.override.domainname.tld (UA string, supports subdomains, too)
general.useragent.site_specific_overrides (true/false)

These two do not work in desktop versions of Firefox.

[...]

PR #884, opened 2019-03-13
Always show URL prefix.

Peacock365:
There is no reason to hide "http://" or "https://" from the user. In fact, it is better for the user to know whether or not the more secure protocol is used.

### DIFF
--- a/browser/app/profile/firefox.js
+++ b/browser/app/profile/firefox.js
@@ -345,7 +345,7 @@ pref("browser.urlbar.maxCharsForSearchSuggestions", 20);
 pref("browser.urlbar.suggest.history.onlyTyped",    false);
 
 pref("browser.urlbar.formatting.enabled", true);
-pref("browser.urlbar.trimURLs", true);
+pref("browser.urlbar.trimURLs", false);
 
 pref("browser.urlbar.oneOffSearches", true);
 


MrAlex94:
It doesn't quite do what you think.

https://developer.mozilla.org/en-US/docs/Mozilla/Preferences/Preference_reference/browser.urlbar.trimURLs

Peacock365:
OK, so it hides only "http://", but not "https://"? I didn't know, as I switched the preference to false right away in my setup. Anyway, URL shortening is bad practice. I know Waterfox inherited this behavior from Firefox, so this is not directed against you. The full address should be displayed, period. Why shorten it?

Peacock365:
As far as I can tell, this change has only upsides, but no downsides. Unless I have missed something.

Peacock365:
@MrAlex94 

Is there any good reason for not merging this? This causes no breakage, this commit would just end the bad practice of URL shortening. No risk here, why not merge it?

WagnerGMD:
> Anyway, URL shortening is bad practice.

That's right @Peacock365 and for a lot of people, it's also a security issue. Basically, the url must be human-readable entirely.

```
lockPref("browser.urlbar.formatting.enabled",false);
lockPref("browser.urlbar.trimURLs",false);
//	https://github.com/MrAlex94/Waterfox/search?q=punycode&type=Issues
//	https://github.com/MrAlex94/Waterfox/issues/261
//	https://github.com/MrAlex94/Waterfox/issues/447
lockPref("network.IDN_show_punycode",true);
```

grahamperrin:
If I understand correctly, 

-1 to locking. 

Allow the user to make his or her own decision.

WagnerGMD:
1) Do you know the definition of "example" ? Apparently @grahamperrin that's not the case.
**Once again, no you are wrong and you didn't take the time to think ! Because that's right, it's just one example.**
2) Did I encourage (or force) the people to follow my decision ? **The answer is absolutely no !**
3) Do you see any of these words such as "advice" etc ? **Damn right, no (there are absent) because it's just an(other) suggestion.**
4) **Perhaps there is an optional purpose ? Guess what, the answer is yes !**
By the way, the people can use a search engine specially to learn a few things about the file **mozilla.cfg** (which is far better for me than the file **prefs.js** (inside the profile)).

Damn it, that's not my trouble if you can't see the difference. Next time let's hope, you will learn and stop this kind of bad behavior and to juge anyone or any reply.
**Learn ? Perhaps the first thing is to ask (one or several questions) before to jump to another bad (or silly) decision.**
From my point of view, I rather to lock these settings because like I said : **it's security issue** (malware, phishing, etc).
```
Pref("browser.urlbar.formatting.enabled",false);
Pref("browser.urlbar.trimURLs",false);
//	https://github.com/MrAlex94/Waterfox/search?q=punycode&type=Issues
//	https://github.com/MrAlex94/Waterfox/issues/261
Pref("network.IDN_show_punycode",true);
--------------------------------------------------------------------------------
--------------------------------------------------------------------------------
defaultPref("browser.urlbar.formatting.enabled",false);
defaultPref("browser.urlbar.trimURLs",false);
//	https://github.com/MrAlex94/Waterfox/search?q=punycode&type=Issues
//	https://github.com/MrAlex94/Waterfox/issues/447
defaultPref("network.IDN_show_punycode",true);
```

Peacock365:
Both @WagnerGMD and @grahamperrin raise valid points. I agree with @WagnerGMD that this is definitely security-related, the whole URL should be shown if possible. But I also agree with @grahamperrin that this setting shouldn't be locked - if people want to hide parts of the URL in their personal setup for whatever reason they should be free to do so. I am in favor of changing the default value of the setting without locking the pref.

That being said, I'm afraid @MrAlex94 won't merge this any time soon despite there not being a good reason for not merging this.

grahamperrin:
> changing the default value of the setting without locking the pref.

Exactly, thanks.

WagnerGMD:
By the past, I already warn you @grahamperrin and this time, I will not close my eyes (neither forget). 
Today it was very fast (you made another bad decision) and it's confirm the fact : you're blind ! 
**You want any help on reddit (etc) ? If you don't stop the down emoji, you will lose it !** I'm not talking about me but the other people. Because unless reddit improve the website, I will stay vanished.

Did you really read my second reply ? **It doesn't seem to be the case at all.** 
Locking ? Did you try at least ? **Let me guess the answer is no, right ? Because the both syntax, prove one thing : the people could change the values of these settings at anytime.**

Does it prove another thing ? Indeed I had give 2 different examples and the only trouble, it's your behavior. You can't just ask one question ? Too hard ? Stop to lie would you ! Because no you made this choice to be disrespectful and it was a big mistake.

How many months did I wait ? [As you can see there](https://github.com/MrAlex94/Waterfox/issues/235) a very very very long time (almost 8 months)...
Then you're right @Peacock365 to be afraid.

MrAlex94:
This hasn't been really much of a priority because:

http://www.example.com/
becomes
www.example.com

Why is this so terrible that needs @WagnerGMD to be so rude?

I'm closing this issue, but if you can show why this actually could pose an issue (maybe an example?) I'm more than happy to concede and flip the pref. 

So far, very few people have mentioned flipping this pref by default, but from what I can tell the handful of people who want it changed can just flip the pref themselves. I could understand if the trimming was really bad and could cause serious phishing issues.

grahamperrin:
Mozilla bug [1462556 - Mark HTTP page as insecure on address bar](https://bugzilla.mozilla.org/show_bug.cgi?id=1462556)

From a POLA perspective of a user of Firefox, someone who may use both browsers (or 'switch' in either direction), it can make sense to not do so (in Waterfox) until Mozilla does so (in a release of Firefox ESR). 

### A comparison with Chromium

image

It's subtle – arguably too subtle – but a static screenshot can not convey the effectiveness. 

Related: [Today’s the day that Chrome brands plain old HTTP “not secure” | Ars Technica](https://arstechnica.com/gadgets/2018/07/todays-the-day-that-chrome-brands-plain-old-http-as-not-secure/) (2018-07-24) in particular the link back in time to the May article. 

Reddit discussions of the Ars Technica articles: 

1. https://www.reddit.com/r/SkydTech/duplicates/7wg5d5/from_july_on_chrome_will_brand_plain_old_http_as/ (no commentary)
2. https://www.reddit.com/r/technology/duplicates/8k7fq4/as_the_web_moves_toward_https_by_default_chrome/
3. https://www.reddit.com/r/SkydTech/duplicates/91jipn/todays_the_day_that_chrome_brands_plain_old_http/ (no commentary)

The URL in the screenshot: 

- <http://web.archive.org/web/20040604050532/http://www.onlamp.com/pub/a/python/2004/06/03/mitch_kapor_interview.html>

Peacock365:
@MrAlex94 

There are several reasons for enabling this, I am going to mention some in the following:

- **Consistency**: Things like "ftp://" or "https://" are being displayed, so it makes sense to also display "http://" for consistency reasons.
- **Copy & Paste:** Some programs don't recognize a hyperlink without the "http://" protocol, treating the address that was copied from the Waterfox's address bar as plain text. Flipping this pref circumvents this issue by default. The possibility of broken links unnecessarily increases if the pref remains as it is now.
- **Best practice behavior:** The whole address should be human-readable, as outlined in the related RFCs. The URL is incomplete without the protocol, URLs without protocols have never been standardized. It also teaches people that the URL begins with "www.", when in fact it doesn't.
- **Security:** HTTP sites are not encrypted, the "http://" protocol is the reason for it. People should be made aware of it as much as possible, flipping the pref contributes to that.

I think my second point is the most important. Some programs, especially older ones, fail to register a hyperlink copied to them when there is no "http://" in front of it. Also, I just don't see a reason to hide it from the user. Hiding it doesn't create _any advantage_ at all, which is why I don't get how @MrAlex94 considers it to be a good idea to have it hidden. 

@MrAlex94, the overwhelming number of responses under this article, for example, is in favor of restoring "http://" to the address bar:

https://www.ghacks.net/2011/09/28/firefox-add-http-back-to-address-bar/

You will also find that most Firefox privacy guides and other about:config recommendations out there recommend this pref to be flipped, for good reason.

MrAlex94:
https://github.com/MrAlex94/Waterfox/commit/08473aef38ecac750042cc70a112047db0007568

> Copy & Paste: Some programs don't recognize a hyperlink without the "http://" protocol, treating the address that was copied from the Waterfox's address bar as plain text. Flipping this pref circumvents this issue by default. The possibility of broken links unnecessarily increases if the pref remains as it is now.

The full URL _is_ copied, not the one shown in the address bar.

Regardless, it's patched now. I don't want to spend any more time on this.

Peacock365:
Thanks for toggling the pref, Alex. 

> Regardless, it's patched now. I don't want to spend any more time on this.

Why so salty? You asked for reasons to toggle the pref, and I mentioned some reasons. I opened this pull request to try and improve the default preferences of Waterfox, thereby helping the project a tiny bit. My contribution was modest, yes, but there is no reason to be angry at me for suggesting a useful change that doesn't negatively impact browsing experience in any way.

grahamperrin:
Interesting: 

1. vote down my suggestion to not lock the preference
2. celebrate the end result (a patch that does not lock the preference). 

It's the weekend. Time for me to celebrate inconsistency and not spend any more time on this :-)

Peacock365:
Just so that you know, @grahamperrin:

You were downvoted by me because you happen to frequently downvote others for no reason at all. Now swallow your own medicine.

grahamperrin:
> no reason 

Rudeness is generally not acceptable. 

Sometimes I ignore rudeness. 

Sometimes I downvote rudeness, and this can be true where: 

- a comment with some (or no) technical merit is unnecessarily rude. 

Sometimes I downvote commentary that is off-topic. 

Sometimes a repository owner will choose to lock something in response to rudeness or other unacceptable behaviour. Here, for example: 

- https://github.com/jasperla/openbsd-wip/issues/86#event-1461028813

– and more recently, where a plea to move discussion was ignored: 

- https://github.com/browser-update/browser-update/issues/397#event-1697538753

PS – to avoid misunderstanding – my down-vote at https://github.com/browser-update/browser-update/issues/397#issuecomment-393540925 was not for rudeness. As far as I recall, it was simple disagreement; explanations may be gained from earlier #178

PR #915, opened 2019-04-11
Fix preprocessor directive in about:license

laniakea64:
Typo was introduced from https://github.com/MrAlex94/Waterfox/pull/781 , someone accidentally deleted a necessary newline.

### DIFF
--- a/toolkit/content/license.html
+++ b/toolkit/content/license.html
@@ -84,7 +84,8 @@
       <li><a href="about:license#fuzz-aldrin">fuzz-aldrin License</a></li>
       <li><a href="about:license#hunspell-nl">Dutch Spellchecking Dictionary License</a></li>
 #if defined(XP_WIN) || defined(XP_LINUX)
-      <li><a href="about:license#twemoji">Twemoji License</a></li>#endif
+      <li><a href="about:license#twemoji">Twemoji License</a></li>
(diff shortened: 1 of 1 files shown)


laniakea64:
Mac OS users: how does Waterfox 56.2.9 `about:license` look for you **without** this patch?  Is it truncated?

laniakea64:
> Mac OS users: how does Waterfox 56.2.9 `about:license` look for you **without** this patch? Is it truncated?

With Waterfox 56.2.9 now out, I was able to extract the released dmg and answer this question myself: **yes it is very truncated.**  As in, the 'table of contents' is mostly missing, and none of the actual license texts are there at all!

@MrAlex94 Given that `about:license` exists for legal reasons, is this patch worth a 56.2.9.1?

laniakea64:
So, to clarify the actual effect of the typo which this pull request corrects:

 - **Mac OS:** `about:license` is very truncated as noted above.
 - **Linux:** there is stray text `#endif` in the middle of `about:license` 'table of contents'.
 - **Windows:** Based on examining the Windows download of 56.2.9, I believe Windows is impacted the same way as Linux.

laniakea64:
Hi @MrAlex94 do you need anything else from me regarding this pull request?

PR #1010, opened 2019-06-07
Add Unity/Global Menu support

hawkeye116477:
Source: `https://bazaar.launchpad.net/~mozillateam/firefox/firefox.xenial/view/1222/debian/patches/unity-menubar.patch`
It can be turned off with `ui.use_unity_menubar` set to false.


MrAlex94:
Thanks 👍

g3ngr33n:
This shouldn't be true by default but per user, turned on.

Issue #979, opened 2019-05-19
topsites entry locked on waterfox alpha


smahat96:
the top sites in the newtab page were missing and also missing from the settings and when I tried to edit "browser.newtabpage.activity-stream.feeds.topsites" in about:config .it was locked !
 and even when editing the pref.js from the profile directory it didn't work .was this on purpose ??

grahamperrin:
Yeah, I don't like to criticise but this type of thing is **not** focused on "[giving users choice](https://www.waterfox.net/about/)": 

![image](https://user-images.githubusercontent.com/192271/57990251-b617a300-7a9c-11e9-972c-c7623da2e2c5.png)

An oversight, maybe? (I can understand; it's an early alpha.)

Or is this a consequence of denying users of Pocket the choice to have saved pages shown in _Highlights_?

smahat96:
could be an oversight and if there is no giude or a way of fixing it myself or I hope it will be fixed in next release.

grahamperrin:
/me looks at the alpha code for the first time 

https://github.com/MrAlex94/Waterfox/blob/f0b4a500c4d0e307866df3172015cc3257263d7e/browser/app/profile/firefox.js#L1200

Speeding lazily through the file history I don't see a point at which it was locked. (Not a problem. Just curious.)

smahat96:
well  since the developer hasn't responded to the issue I have no idea if he saw and if he is going to change it in next release ..i guess we have to wait and see

laniakea64:
I briefly looked into some aspects of the new tab page in v68, including this, and I don't see what's so anti-user about "top sites" feature that it deserves to be hard-disabled.  All I noticed when unlocking the pref and enabling it was the addition of some links to sites + image tiles for said links, as well as one-click access to some installed search engines.  And the sites/search engines that are displayed can be easily customised, AFAICT nothing is being forced or pushed.

This looks useful and again I don't see how this feature is anti-user.  What am I missing?

Hi @MrAlex94 Did you intend to lock `browser.newtabpage.activity-stream.feeds.section.topstories` ("Recommended by Pocket") instead?

laniakea64:
> Speeding lazily through the file history I don't see a point at which it was locked.

Apparently top sites is "privacy-unfriendly" - https://github.com/MrAlex94/Waterfox/commit/1983d09d4f35d6d9e58f8088b55018e40525feb4#diff-d7a0401e94f68887594ad1d1d9f9c28dR1182

[...]

PR #1016, opened 2019-06-08
v68: privacy-ish fixes

laniakea64:
**Disable DiscoveryStream / "Recommended by Pocket"**

Waterfox already crippled this, but it was still active.  Disabling it reduces browser console noise and cleans up the new tab page.

**Disable urlbar speculative connect by default**

`browser.urlbar.speculativeConnect.enabled` when set to `true` causes Waterfox to make weird connections that are difficult to track down.  Standard troubleshooting an affected profile erroneously points to `places.sqlite` as the cause, which results in a wild goose chase.  This should not be enabled unless the user explicitly wants it and knows about it.

**Expose media.gmp-manager.updateEnabled in about:config by default**

Same patch as https://github.com/MrAlex94/Waterfox/pull/478


laniakea64:
Hi @MrAlex94 do you need anything else from me in order to review and merge this?

grahamperrin:
Is it possible to disable things without _locking_ preferences?

laniakea64:
@grahamperrin Yes, that's exactly what this does for urlbar speculative connect.

grahamperrin:
Well, yes, but I'm looking at what's locked ;-)

Consider e.g. #979

laniakea64:
@grahamperrin That's different.  Top sites actually works fine when unlocked and enabled (and I do have a local patch to unlock that pref, and I currently have it flipped on in my daily-use profile).  And that locking was erroneous in the first place - https://github.com/MrAlex94/Waterfox/issues/979#issuecomment-500114917
(and it has been unlocked as I type this :smiley: )

The prefs locked by this PR control something that **doesn't** work and _intentionally_ doesn't work.

Why do you want to enable it?

grahamperrin:
Pocket: because I'm an active user, I very often take a selection of recommendations.

laniakea64:
Hmm, removal of Pocket is a [feature of Waterfox](https://github.com/MrAlex94/Waterfox#features) and "Recommended by Pocket" doesn't work in Waterfox.  And I've found "Recommended by Pocket" to be somewhat pushy in Firefox and not so easy to disable there.  So I would rather see this PR merged as-is.

Maybe you could raise a separate issue requesting to add a mozconfig option to enable Pocket and "Recommended by Pocket" at build time?  Such mozconfig option would mean these prefs would be locked to `false` in normal Waterfox builds, but unlocked in Pocket-enabled builds, and this way Pocket stuff would still not be forced on users who don't want it.

MrAlex94:
All looked good to me. Sorry had received this just as I set off traveling!

Issue #1090, opened 2019-08-04
68.0a2 tabs now have black background with black text


laniakea64:
**Describe the bug**
In latest 68.0a2 tabs background is black with Default or Light theme.  Because those themes use black text, this leaves the tab unreadable and visually unusable.

**To Reproduce**
Tools > Add-ons Manager > Themes, enable either Default or Light

**Expected behavior**
Tabs should be gray/light-gray with black text.

**Screenshots**
![Screen shot 2019-08-04 1](https://user-images.githubusercontent.com/16628411/62426178-8e090a80-b6af-11e9-8f61-79d3a65dd131.png)

**Desktop (please complete the following information):**
 - OS: Xubuntu 18.04
 - Waterfox 68.0a2 self built from https://github.com/MrAlex94/Waterfox/commit/b5059dc5ab2699d335dfa9c8c01a5a3c3a7942bd

**Additional context**
Sometimes the tabs look as expected briefly after a restart.  But as the browser is used, e.g. if the Add-ons Manager is opened, it will revert back to the unusable black background.

The black background also happens with Dark theme and with personas, but in those cases the text is white and light grey respectively, so one workaround could be to use either Dark theme or a persona.

grahamperrin:
68.0a2 on Windows (to the right), default theme: 

![image](https://user-images.githubusercontent.com/192271/62426730-7433eb80-b6e0-11e9-87a8-db56fa20ba88.png)

– and light: 

![image](https://user-images.githubusercontent.com/192271/62426738-99285e80-b6e0-11e9-81ad-46029400d52a.png)

grahamperrin:
Reproducible with (unannounced) build 20190906171246 of 68.0b1: 

![image](https://user-images.githubusercontent.com/192271/64485018-7ffc5b80-d212-11e9-85d8-941adf4ff74a.png)

Below, the bug is **not immediately apparent** – it occurs (for example): 

- after taking the about:preferences route to **about:support**

https://s.put.re/pbN6Ux85.mp4

Another example, thanks to @troysjanda:

- after taking the hamburger menu route to **about:addons**

… and so on.

grahamperrin:
Cross references: 

- https://www.reddit.com/comments/d15yea/-/ezr7139/?context=1
- [UserChrome.css stops working if user open about:addons : waterfox](https://www.reddit.com/r/waterfox/comments/d1z27a/userchromecss_stops_working_if_user_open/)

[...]

Issue #459, opened 2018-03-03
Devtools dark theme makes scrollbars invisible


laniakea64:
Waterfox 56.0.4.1, new profile
Xubuntu 16.04 64-bit
Greybird GTK theme

**Steps to reproduce:**

1) on the Waterfox blog page that opens at startup, open the devtools
2) open the Style Editor.  Notice there are vertical and horizontal scrollbars.
3) Open devtools settings and change the devtools theme to Dark
4) Go back to the Style Editor

**Expected results:** vertical and horizontal scrollbars visible in the Style Editor

**Actual results:** Style Editor's vertical and horizontal scrollbars are invisible.  They are still there though - clicking/dragging the mouse where the scrollbars should be, does produce expected scrollbar behavior.

grahamperrin:
> … Waterfox blog page …

Can you reproduce the issue elsewhere? 

There may be something unusual about that page, it features in a bug that involves an experimental legacy extension: 

- [Scrollmap 0.9 disappears after loading www.waterfoxproject.org/blog/ · Issue #15 · vovcacik/scrollmap](https://github.com/vovcacik/scrollmap/issues/15)

– as far as I can tell, not an issue with any Waterfox blog page at the staging site.

laniakea64:
Yes, I can reproduce it no matter what page I'm on when I open devtools.

In case it wasn't clear, this affects **all** scrollbars in the devtools, and only the scrollbars within devtools.  Other scrollbars are not affected.

grahamperrin:
Sorry! I misread. You were clear. I thumbed down my own comment :-)

jbeich:
Can you attach a screenshot? Maybe you didn't look [hard enough](https://bugzilla.mozilla.org/show_bug.cgi?id=1100622).

laniakea64:
Screen shot of light theme -
![waterfox-devtools-light](https://user-images.githubusercontent.com/16628411/36942232-59a0571a-1f3b-11e8-96e3-6d945aa2f85e.png)

Screen shot of dark theme -
![waterfox-devtools-dark](https://user-images.githubusercontent.com/16628411/36942233-643c0ade-1f3b-11e8-8d98-5989a76fec73.png)

laniakea64:
It's GTK theme dependent.  With Adwaita GTK theme the scrollbars are visible, although hard to see.

laniakea64:
[...]

PR #1112, opened 2019-08-27
v68: fixes for the Australis-Photon themes

laniakea64:
**Get the Australis-Photon themes working as intended and fix consistency issues**

Fixes https://github.com/MrAlex94/Waterfox/issues/1090

Apparently Waterfox had already added several distinct Australis-Photon styles, one for each built-in theme, but only the dark Australis-Photon style was ever getting applied.  This patch introduces selecting the correct Australis-Photon style for the selected built-in theme, and fixes the consistency issues applying it.  Seems to work well as long as the user does not manually mess with `extensions.activeThemeID`.  Given how it was implemented before this patch, I assume that edge case is not a concern.

Note: Seeing the lack of an `else` clause in the existing implementation, I assumed the design intent was to not apply any Australis-Photon style when a persona is in use.

**Add a pref to disable the Australis-Photon themes**

See https://github.com/MrAlex94/Waterfox/issues/1131 for context.


laniakea64:
Hi @MrAlex94 did I correctly understand the intent?
Do you need anything else from me on this in order to review and merge?

devarped:
I've built Waterfox with that commitand I can see that tabls use colors from Light theme on Default theme, which in my opinion does not look great.
tabs

laniakea64:
> I've built Waterfox with that commit and I can see that tabls use colors from Light theme on Default theme, 

Good catch, I didn't notice that `australis-default.css` was just a copy of `australis-light.css`!  This is not right, it means the Australis-Photon Default theme will make unreadable white-text-on-white-tabs for users with dark system theme (and I have confirmed this with Adwaita-Dark GTK theme).

I was able to fix `australis-default.css` to make active tab color based on system theme so it doesn't create unreadable text, but I'm not sure this fix is suitable for a PR, since with some GTK themes it makes the active tab not so easy to distinguish from the tab-bar background.  It would also need to be tested on Windows and Mac OS, which I'm not able to do.

If you'd like to take a look and experiment with it, I pushed it to https://github.com/laniakea64/Waterfox/tree/v68-theme-fix-2

devarped:
>I'm not sure this fix is suitable for a PR, since with some GTK themes it makes the active tab not so easy to distinguish from the tab-bar background.

Maybe add some kind of thin border to the active tab? It will definely change previous style of tabs, but I don't think it will look too bad.

laniakea64:
> Maybe add some kind of thin border to the active tab? It will definely change previous style of tabs, but I don't think it will look too bad.

That's how Waterfox 56 default theme solves this, see e.g. screenshots in https://github.com/MrAlex94/Waterfox/issues/459#issuecomment-370201445

Unfortunately I don't know enough about SVG to know how to add the borders, sorry.

laniakea64:
The Australis-Photon themes are overriding people's userChrome.css, and this patch is going to make that breakage even more prevalent, see https://github.com/MrAlex94/Waterfox/issues/1131#issuecomment-530021496

@grahamperrin suggested that the adding a `about:config` pref to disable the Australis-Photon themes would be the best solution for that issue - https://github.com/MrAlex94/Waterfox/issues/1131#issuecomment-530068924
I would like to add such patch to this PR before merge.  (I will keep the Australis-Photon themes enabled by default.)

laniakea64:
> I would like to add such patch to this PR before merge. (I will keep the Australis-Photon themes enabled by default.)

Done.

Note to testers: this seems to require once starting Waterfox with `-purgecaches` command-line switch.

MrAlex94:
Okay so my plan is to have a separate and default Australis-Photon theme, and keep the other defaults as well for those who want to use custom CSS. That way, we keep the Waterfox look while letting people customise as they please. Sorry for not making this clearer. I'll leave this here for now and come back to review if anything changes.

laniakea64:
Thanks @MrAlex94 for responding.

I'm not sure if I will be able to find more time to look into this, but in case I can:

So the files for Dark/Light themes are at https://github.com/MrAlex94/Waterfox/tree/gecko68/browser/themes/addons , but where are the equivalent files for Default theme?

MrAlex94:
I believe it's spread throughout the https://github.com/MrAlex94/Waterfox/tree/gecko68/browser/themes directory, it's not packaged like the compact themes.

grahamperrin:
Thanks folks and @MrAlex94 to further clarify, is the plan to have **a GUI to a preference** e.g. https://github.com/MrAlex94/Waterfox/pull/1112/commits/adc456d4f724098dca6260d904778fb63b5d0cd9?

laniakea64:
@grahamperrin Seems clear to me, that the plan is to make it separate entries in Tools > Add-ons > Themes, with one of the Australis-Photon themes being default, and the option of straight Photon offered there in the same way Light and Dark themes are now.

PR #1158, opened 2019-09-18
(WF68) Enable containers by default.

Peacock365:
See header.

### DIFF
--- a/browser/app/profile/firefox.js
+++ b/browser/app/profile/firefox.js
@@ -1642,8 +1642,8 @@ pref("privacy.userContext.ui.enabled", true);
 // 0 disables long press, 1 when clicked, the menu is shown, 2 the menu is shown after X milliseconds.
 pref("privacy.userContext.longPressBehavior", 2);
 #else
-pref("privacy.userContext.enabled", false);
-pref("privacy.userContext.ui.enabled", false);
+pref("privacy.userContext.enabled", true);
+pref("privacy.userContext.ui.enabled", true);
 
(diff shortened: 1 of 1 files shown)


laniakea64:
This also needs `privacy.userContext.longPressBehavior` set to `2`

Ibuprophen:
@Peacock365, I did see what @laniakea64 may have been referring to... 

`Line 1643` refers to `privacy.userContext.longPressBehavior`being set to `2`but, `Line 1649` reflects `0`.


~Ibuprophen

grahamperrin:
https://blog.mozilla.org/tanvi/2017/10/03/update-firefox-containers/#comment-36309

Need to check the (Mozilla) rationale for continuing to not enable these features by default. IIRC there was an overriding wish to promote extensions-based control of the preferences but a long time has passed since then …

laniakea64:
> Need to check the (Mozilla) rationale for continuing to not enable these features by default. 

I looked into this a while ago, and IIRC it boiled down to, they put more features in their Multi-Account Containers extension and prefer people use that.  See e.g. at the bottom of https://support.mozilla.org/en-US/kb/containers

Edit: And also from the [description of Multi-Account Containers](https://addons.mozilla.org/en-US/firefox/addon/multi-account-containers/):
> Firefox Multi-Account Containers was first introduced as "Containers" available only in Firefox Nightly. It went on to be a Test Pilot Experiment, where we improved the user experience and added new features. Now it exists here as an extension that can be installed by all Firefox users.
> 
> Although you can still use the feature built into Firefox Nightly, we recommend using this extension as it has a richer user experience.

Issue #1079, opened 2019-07-24
Waterfox 56 rendering issue on many sites


N-R-G:
Various site aren't correctly visualized with waterfox 56.2.12

example:
https://lichess.org/editor/8/8/8/8/8/8/8/8_w_-_-
appears to be an issue only with old firefox forks (before quantum):
https://awau.moe/4F5g5J3.png

others example in this reddit thread:

https://www.reddit.com/r/waterfox/comments/cfbh47/waterfox_breaks_certain_pages_on_pimoronicom/

hawkeye116477:
Possibly related => https://bugzilla.mozilla.org/show_bug.cgi?id=1302667

Update: One of error on console is about `worker-src`, but after adding changes from `https://hg.mozilla.org/mozilla-central/rev/bbb4d0097263` and building rendering bug still happens, only error on console disappears .
Seems that something is not right with `grid-area: spare-bottom;`.

Issue #1066, opened 2019-07-15
waterfox can't display images with div and a elements


vesselict:
I have problems with latest version on both windows and linux of displaying images on sites without `<img />` element but with` <div>` and `<a> `elements ...

On firefox everything works just fine.

laniakea64:
> latest version

56.2.12 or 68.0a2?



> sites without `<img />` element but with` <div>` and `<a> `elements ...

Could you please share some example links?

vesselict:
sure..
https://www.insomnia.gr/classifieds/category/68-motorola/

56.2.12 I had this problem on older versions too.

laniakea64:
I see the same issue here in Waterfox 56.2.12, new profile.  Works fine in Waterfox 68.0a2 in my main profile.

So far I was unable to figure out why it isn't working in 56.2.12.  There are no related errors in the console, and setting user-agent override of Firefox 60 or Firefox 68 didn't help.

That site is using a standard method of 'lazy loading images', i.e. the images should load only when you scroll them into view, so I would suspect this breakage is not isolated to the one site.

colvacdir:
As of version 56.2.14, this problem still exists. Has any progress been made on what's causing the issue and if it will be fixed in a future version?

I tested the sample URL mentioned above with a couple of similar browsers, Firefox ESR 52.9.0 and TenFourFox FPR 16, and it works fine with both of those, which indicates it's specifically tied to Waterfox.

hawkeye116477:
It seems that images are loaded by `https://cdn2.insomnia.gr/uploads/javascript_global/root_framework.js.76eb380d05b2cd352f3268aa6cf46ad8.js`. It uses `data-background-src` to apply background to css. That's fixed on FF 57.

micokeman:
I also noticed that this site deosn't work on Waterfox version 56.2.14 

https://www.mlb.com/standings#wildCardSubNav

grahamperrin:
Meta, tracking: #538

@micokeman FYI there's related discussion at [mlb.com Banner Rendering Issues. : waterfox](https://www.reddit.com/r/waterfox/comments/d62qit/mlbcom_banner_rendering_issues/)

hawkeye116477:
**Mozregression result:**
- Last bad: Firefox v57.0a1 (2017-09-18)
- First good: Firefox v57.01 (2017-09-19)

[...]

Issue #1065, opened 2019-07-15
Waterfox does not display user profile info properly on QRZ.com with uBlock Origin installed


willbprog127:
**Describe the bug**
Amateur radio operator profile information does not show up in Waterfox when using uBlock Origin on QRZ.com.  This is with a completely new Waterfox profile.  Using a completely new profile in Firefox with uBlock Origin works fine, so it does not appear to be only a uBlock Origin problem.

**To Reproduce**
1. Rename your current Waterfox profile to something else (just for testing)
2. Start Waterfox
3. Go to the Firefox add-ons page and install uBlock Origin
4. Go to https://www.qrz.com/db/N6ZN
5. Observe that there is *no* user profile information for N6ZN under the 'Biography' tab, even though the footer for that section shows 1067 bytes currently being used for the Biography (meaning it's not an empty biography).

**Expected behavior**
N6ZN's biography information should be displayed.

**Desktop info**
 - OS: macOS 10.14.x, MacOS 10.11.x
 - Waterfox version: 56.2.12 (64-bit)

**Additional context**
Although I'm not 100 percent sure, it appears this site is using Javascript and JQuery to synthesize an iframe to display the user's information.  This works fine on Firefox with uBO but obviously not on the current Waterfox.

This bug is worrisome, because what if other content on other sites is not being displayed due to this bug?  I do a lot of research, especially for computer programming information, and this could omit important information I need.

Thanks! 👍

grahamperrin:
Also [Issue with Water fox incorrectly loading QRZ.com · Issue #612 · MrAlex94/Waterfox](https://github.com/MrAlex94/Waterfox/issues/612)

hawkeye116477:
**Mozregression result:**
* Last bad: Firefox v58.0a1 (2017-09-21)
 * First good: Firefox v58.01 (2017-09-22)

Pushlog:
https://hg.mozilla.org/mozilla-central/pushloghtml?fromchange=319a34bea9e4f3459886b5b9e835bd338320f1fd&tochange=3d72fdb0e561ea59d9e5850c3e71367dbb8a7148

Update: I guess that's the fix => https://bugzilla.mozilla.org/show_bug.cgi?id=1379148, I'll test and add to PR.

grahamperrin:
@hawkeye116477 OT, please, why do you sometimes suppress links?

[...]

Issue #1004, opened 2019-06-02
Cursor disappears, particularly on Discourse forums


colinl:
**Describe the bug**
After hitting Reply on a forum using Discourse (https://www.discourse.org/) the cursor does not appear in the edit window.  Typing works but the cursor is not visible.  Clicking away from Waterfox and back shows the cursor, though it disappears after a while though I have not tied down exactly what makes it disappear.

**To Reproduce**
Steps to reproduce the behaviour:
1. Open a topic on a discourse forum to which one is logged in .
2. Click Reply at the bottom of the topic
3. The new post field opens but there is no cursor visible.

**Expected behavior**
A cursor should appear

**Desktop (please complete the following information):**
 - OS: Ubuntu
 - Version 18.10

**Additional context**
Waterfox 56.2.10 (64-bit)

I see this effect on both the node-red and grafana forums (https://discourse.nodered.org and https://community.grafana.com) as well as the discourse demo forum https://try.discourse.org/
Unfortunately I have not been able to find a forum that does not require logging in in order to reply so that others can more easily confirm this, however one can either create and account at the demo forum at https://try.discourse.org/ or sign in with social media.
The only report of note in the web console is 
`This site appears to use a scroll-linked positioning effect. This may not work well with asynchronous panning; see https://developer.mozilla.org/docs/Mozilla/Performance/ScrollLinkedEffects for further details and to join the discussion on related tools and features!`
but I suspect that is not related.

Firefox 67.0 does not show the problem.

grahamperrin:
> Clicking away from Waterfox and back shows the cursor,

When I lose sight of the pointer: 

1. <kbd>Alt</kbd>-<kbd>Tab</kbd>
2. <kbd>Alt</kbd>-<kbd>Tab</kbd>

KDE Plasma, FreeBSD-CURRENT. 

I see the symptom maybe a few times per week, but I've never paid enough attention to tell whether the issue is more likely to occur with Waterfox than with other applications. From now on, I'll pay attention.

grahamperrin:
PS sorry, now I see the issue – **the cursor**, not the pointer. 

Consistently reproducible with Waterfox Classic on FreeBSD-CURRENT, not reproducible with 68.0a1. 

Mozilla Discourse can be used with a Firefox account. 

1. https://discourse.mozilla.org/
2. view any topic that's open to replies
3. either (a) click **Reply** or (b) select part of a post then click **Quote**

Bonus: https://discourse.mozilla.org/read

[...]

PR #1164, opened 2019-09-19
Add GtkFileChooserNative support, new js properties and other improvements to Waterfox Classic

hawkeye116477:
Borrowed from https://github.com/MoonchildProductions/UXP/commit/47c48dfcb87a574f6b96f3abf3ed14b9b08b2371.
This adds support for native KDE dialog.
But to make it really working, users needs to have `xdg-desktop-portal-kde` package and run Waterfox with `GTK_USE_DIALOG=1`.
That package is available from Ubuntu 18 and Debian 10. But on GTK lower than 3.24.4 can be some problems (Ubuntu 18), so in that case SUSE's solution which I use on my package still will be better (even it works on Ubuntu 16 and Debian 9).


grahamperrin:
Thanks! 

<strike>https://lists.freebsd.org/pipermail/freebsd-current/2019-September/074405.html</strike>

https://lists.freebsd.org/pipermail/freebsd-current/2019-September/074442.html

> … dbus portal …

hawkeye116477:
Additionally:
* Added new js properties => `Symbol​.prototype​.description, String.prototype.trim{Start,End}`.
* Fix for emoji (#919). Still some emojis are not colored, but now it's better, they look like same as on newest Firefox/Waterfox Current and is possible to change font for emojis.
* Experimental support for AV1 in MP4 container (#897, disabled by default, can be enabled by setting media.av1.enabled to true in about:config, also to make it working needed is adding `ac_add_options --enable-av1` to mozconfig). I added similar things which are on UXP, so may be same problems as on https://github.com/MoonchildProductions/UXP/issues/888.
* Small fixes for MP4 parser.
* Added the path to the running binary to about:support. 
* Added new strings for language packs (#903 and CSP things which I added previously, only Polish language is translated, other are in English).
* Fixes for WebP (some changes borrowed from UXP/Basilisk)

I also created repository with unpacked language packs for easier collaboration => `https://github.com/hawkeye116477/LanguagePacksForWaterfox` :smile:. I think that repository should be created by @MrAlex94, Cyberfox and Pale Moon also has similar.

Update: I added also some other useful fixes.

MrAlex94:
Thanks. I'm going to change the way language packs are handled, and do it at the packing stage instead. Will go through these all, cheers for the commits!

hawkeye116477:
@MrAlex94 Ok. I added also fixes for #612 and #1065 now. You can merge, all I tested and seems that everything works as intended :smile:

Update: I also added fix for #1079 (only problem with Nintendo mentioned on Reddit not solved).

Ibuprophen (commented):
Looks Very Good!

Kudos to You @hawkeye116477!!! :-) 


~Ibuprophen

hawkeye116477:
I added some mp4 crash fixes and fix for `mediamarkt.de` (https://www.reddit.com/r/waterfox/comments/c38byn/page_layout_problem/) now.

grahamperrin:
☝ thumbs up for sheer effort! (I haven't tested, but don't doubt the goodness of all of this …)

hawkeye116477:
Now added fix for #1066 and https://www.reddit.com/r/waterfox/comments/d62qit/mlbcom_banner_rendering_issues/.

@MrAlex94 Do I need to do something to green light this :smile:? And don't worry, I left Nintendo's site for you (Shadow DOM v1) and maybe `https://bugzilla.mozilla.org/show_bug.cgi?id=1392177`.

Updated: Added fix for #1004 also.

MrAlex94:
> @MrAlex94 Do I need to do something to green light this 😄? And don't worry, I left Nintendo's site for you (Shadow DOM v1) and maybe `https://bugzilla.mozilla.org/show_bug.cgi?id=1392177`.

Not at all! Just testing the features that have been affected by the changes on all the platforms, just in case there's a breakage somewhere. Has happened to me so many times before when I merge changes from upstream, trying to be a bit careful this time so I don't ruin things for everyone 😅. Sorry it's taking so long, literally testing as we speak, VMs are slow!

hawkeye116477:
Ok, all in all right, if you're in a hurry, the devil is happy.

MrAlex94:
Excellent job! All worked perfectly.

PR #1170, opened 2019-09-25
(WF68) Disable battery API, anti-fingerprinting measure.

Peacock365:
See header. Align the behavior of Waterfox 68 with Waterfox 56.


grahamperrin:
https://developer.mozilla.org/docs/Web/API/Battery_Status_API

> **Deprecated**
> 
> This feature is no longer recommended. …

PR #1193, opened 2019-10-13
(WF68) Enable the Awesome Bar by default.

Peacock365:
Reason: The Awesome Bar uses XUL / XBL, while the Quantum Bar uses web technologies. The Awesome Bar is desirable for Waterfox 68 / Current. Also, it is similar to the Waterfox 56 address bar. Not locking this pref, one may revert to the new Quantum Bar if desired. No difference in functionality.

### DIFF
--- a/browser/app/profile/firefox.js
+++ b/browser/app/profile/firefox.js
@@ -346,7 +346,7 @@ pref("browser.urlbar.usepreloadedtopurls.enabled", false);
 pref("browser.urlbar.usepreloadedtopurls.expire_days", 14);
 
 // Enable the new Address Bar code.
-pref("browser.urlbar.quantumbar", true);
+pref("browser.urlbar.quantumbar", false);
 
 pref("browser.altClickSave", false);
 


laniakea64:
> Reason: The Awesome Bar uses XUL / XBL, while the Quantum Bar uses web technologies. The Awesome Bar is desirable for Waterfox 68 / Current. 

I'm not following this reasoning.  Could you please explain more?

> Also, it is similar to the Waterfox 56 address bar. 

What is the user-visible difference between Waterfox Current Awesome Bar and Quantum Bar?  I've played with toggling this pref before and couldn't tell any difference.

Peacock365:
@laniakea64 

The Awesome using XUL / XBL might facilitate porting pre-Firefox 57 themes to Waterfox 68. There are cosmetic differences in the drop-down menu: The new Quantum Bar no longer contains the option to select a different search engine directly from the address bar drop-down menu. Also, it has removed the shortcut for search engine management.

laniakea64:
> The Awesome using XUL / XBL might facilitate porting pre-Firefox 57 themes to Waterfox 68. 

Thanks, this makes sense.

> There are cosmetic differences in the drop-down menu: The new Quantum Bar no longer contains the option to select a different search engine directly from the address bar drop-down menu. Also, it has removed the shortcut for search engine management.

For me this is controlled by `browser.urlbar.oneOffSearches`, and seems to work in both Quantum Bar and Awesome Bar.

PR #1194, opened 2019-10-13
(WF68) Disable the HTML about:addons page by default.

Peacock365:
Reason: The new design requires more clicks to achieve the same results, the former about:addons page is similar to the one in Waterfox 56, the former about:addons page is also desirable for Waterfox 68 since it is based on XUL / XBL instead of HTML. Not locking this pref, one may revert to the new HTML about:addons page if desired. No difference in functionality.

### DIFF
--- a/modules/libpref/init/all.js
+++ b/modules/libpref/init/all.js
@@ -5204,7 +5204,7 @@ pref("extensions.webextensions.enablePerformanceCounters", true);
 pref("extensions.webextensions.performanceCountersMaxAge", 5000);
 
 // The HTML about:addons page.
-pref("extensions.htmlaboutaddons.enabled", true);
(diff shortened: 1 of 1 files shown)


grahamperrin:
For (distant) future reference:

- Mozilla bug [1564525 - Removing extensions.htmlaboutaddons.enabled pref](https://bugzilla.mozilla.org/show_bug.cgi?id=1564525)

Peacock365:
@grahamperrin 

I am aware of this. By the next Firefox ESR version (Firefox 78) the old add-on manager will be removed completely anyway. Has no bearing on Firefox 68 ESR though, which Waterfox Current is based on.

MrAlex94 (approved):

PR #1195, opened 2019-10-15
(WF68) Various about:config tweaks.

Peacock365:
See header. Various small tweaks, among others the "Not Secure" text in the address bar is now enabled for domains prefixed with "http://".

### DIFF
--- a/browser/app/profile/firefox.js
+++ b/browser/app/profile/firefox.js
@@ -47,7 +47,7 @@ pref("extensions.getAddons.langpacks.url", "", locked);
 pref("extensions.getAddons.discovery.api_url", "", locked);
 
 // Enable the HTML-based discovery panel at about:addons.
-pref("extensions.htmlaboutaddons.discover.enabled", false);
+pref("extensions.htmlaboutaddons.discover.enabled", false, locked);
 
 // The URL for the privacy policy related to recommended extensions.
 pref("extensions.recommendations.privacyPolicyUrl", "https://www.mozilla.org/privacy/firefox/?utm_source=firefox-browser&utm_medium=firefox-browser&utm_content=privacy-policy-link#addons");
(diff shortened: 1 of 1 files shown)


browser/app/profile/firefox.js
```diff
@@ -47,7 +47,7 @@ pref("extensions.getAddons.langpacks.url", "", locked);
 pref("extensions.getAddons.discovery.api_url", "", locked);
 
 // Enable the HTML-based discovery panel at about:addons.
-pref("extensions.htmlaboutaddons.discover.enabled", false);
+pref("extensions.htmlaboutaddons.discover.enabled", false, locked);
```
grahamperrin, line 50: Why lock?

browser/app/profile/firefox.js
```diff
@@ -1865,8 +1865,8 @@ pref("toolkit.coverage.enabled", false, locked);
 pref("toolkit.coverage.endpoint.base", "");
 
 // Discovery prefs
-pref("browser.discovery.enabled", false);
-pref("browser.discovery.containers.enabled", true);
+pref("browser.discovery.enabled", false, locked);
+pref("browser.discovery.containers.enabled", false, locked);
```
grahamperrin, line 1869: Why lock?

laniakea64: @grahamperrin `browser.discovery.enabled` has to do with personalising recommendations.  And that is also disabled by Waterfox hard-disabling health report telemetry.  IMO locking this pref to `false` seems appropriate since it can't do anything else in Waterfox.

See related code -
https://github.com/MrAlex94/Waterfox/blob/19fe621225c69b3a8df8f60cc5d68bd0422601e4/browser/modules/Discovery.jsm
https://github.com/MrAlex94/Waterfox/blob/19fe621225c69b3a8df8f60cc5d68bd0422601e4/toolkit/mozapps/extensions/content/extensions.js#L2343-L2353

laniakea64:
@Peacock365 Why change `browser.uitour.requireSecure`?  Locking that to `false` would seem to only make Waterfox less secure, albeit in a way that's related to a disabled component -
https://github.com/MrAlex94/Waterfox/blob/19fe621225c69b3a8df8f60cc5d68bd0422601e4/browser/components/uitour/UITourChild.jsm#L56-L68

laniakea64:
> Why change `browser.uitour.requireSecure`? Locking that to `false`

I would still like to understand this change, in particular why it is not the security problem it looks like, because right now I'm not sure how comfortable I am with this.  Thanks for any clarification.

Peacock365:
@laniakea64 

_browser.uitour.enabled_ is set to false anyway, it is the meta pref. No connection will be established, since the UI Tour is deactivated in the first place. I did flip the preference you mentioned for publicity's sake, because some alu hats will claim that parts of the UI Tour are still active while in reality they are not.

laniakea64:
Thanks @Peacock365 , that addresses most of my concerns :+1: I'm still not clear why it is _locked_ to `false`, rather than simply _defaulted_ to `false`?

PR #1207, opened 2019-10-17
[current] Disable some telemetry settings

laniakea64:
In https://old.reddit.com/r/waterfox/comments/dib6iq/why_is_the_pref_for_telemetry_locked_to_enabled/ it was pointed out that `toolkit.telemetry.enabled` is locked to `true` in Waterfox Current.  I checked and saw the same, so I looked into this, and it turns out Waterfox was indeed collecting telemetry (only locally, not sending it anywhere).

This patch flips the value of `toolkit.telemetry.enabled` to `false` along with a couple other internal telemetry settings I found while investigating.  And based on comparing `about:telemetry` in builds before and after this patch, I think this also disables telemetry collection.

### DIFF
--- a/modules/libpref/Preferences.cpp
+++ b/modules/libpref/Preferences.cpp
@@ -3334,18 +3334,10 @@ static Maybe<bool> TelemetryPrefValue() {
     return Nothing();
   }
 
-  // Determine the correct default for toolkit.telemetry.enabled. If this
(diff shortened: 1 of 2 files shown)


laniakea64:
Hi @MrAlex94 do you need anything else from me in order to review and merge this?

BTW I'm noticing that Waterfox seems to feel significantly faster with this patch than without.

MrAlex94 (approved):

PR #1213, opened 2019-10-19
Few minor improvements for Waterfox Current

hawkeye116477:
Langpacks are from `http://ftp.mozilla.org/pub/firefox/releases/68.1.0esr/linux-x86_64/xpi/` and little modified by me (changed Firefox to Waterfox, etc => https://github.com/hawkeye116477/LanguagePacksForWaterfox/commit/957efac0c4059e1254f2a40f950cbaa4321c03f3). For langpacks we may need maybe some makefile to put them in correct dir, I tried to add them in `browser/features`, but if they are on that dir, seems not working, however they work in `browser/extensions`, but are disabled and can't be enabled until I switch `extensions.autoDisableScopes` to `3`.

I also added Unity/Global Menu support made by Canonical/Ubuntu Mozilla Team (https://bazaar.launchpad.net/~mozillateam/firefox/firefox.xenial/view/1347/debian/patches/unity-menubar.patch) and path to the running binary to `about:support`.

PR #1223, opened 2019-10-25
[current] Cleanup prefs for "Recommend (extensions/features) as you browse"

### LINKED ISSUE #1058: v68: Disable "Recommend (extensions/features) as you browse"
Currently Waterfox 68 has the following prefs defaulted to `true` -

`browser.newtabpage.activity-stream.asrouter.userprefs.cfr.addons`
`browser.newtabpage.activity-stream.asrouter.userprefs.cfr.features`

These are the settings "Recommend extensions as you browse" and "Recommend features as you browse", respectively.  I suspect these settings might not work in Waterfox, because I didn't notice this until I went through `about:preferences` for unrelated reasons, and the associated "Learn more" links do nothing.  If however these settings actually are functional, and if some users want this, it should be **opt-in**: pushing extensions on people who didn't request such recommendations seems spammy and anti-user.

Both prefs should be defaulted to `false`.  If they don't work they should be locked to `false` to avoid mis-impressions.

laniakea64:
Fixes https://github.com/MrAlex94/Waterfox/issues/1058


hawkeye116477:
I edited main.xul, main.js and firefox.js on #1213. I hope that doesn't cause conflicts.

laniakea64:
> I edited main.xul, main.js and firefox.js on #1213. I hope that doesn't cause conflicts.

It looks like these two PRs touch entirely different parts of those files, so it should be fine, shouldn't it?

laniakea64:
Hi @MrAlex94 did I understand your intent correctly when I looked into this in https://github.com/MrAlex94/Waterfox/issues/1058#issuecomment-546126253 ?  Do you need anything else from me in order to review and merge this PR?

PR #1289, opened 2019-12-05
[Current] Add another improvements and fixes (part 2)

hawkeye116477:
I noticed some problems when changing Tab Bar position, so here are fixes (only tested on Linux).

Now titlebar buttons should stay on top while moving Tab Bar just like on Vivaldi. There was also problem that after enabling menu, titlebar buttons were duplicated, now fixed.

### DIFF
--- a/browser/app/profile/firefox.js
+++ b/browser/app/profile/firefox.js
@@ -1928,3 +1928,6 @@ pref("browser.preferences.useAllNew", false);
 
 // Disable extensions recommendations
 pref("extensions.getAddons.showPane", false);
+
+// Window controls position
+pref("browser.windowControls.position", "right");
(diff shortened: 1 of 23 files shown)


hawkeye116477:
@MrAlex94 Are you tested that and had any problems?

hawkeye116477:
Additional new changes:
* added possibility to set Window Controls position
* added possibility to set Waterfox Current logo as menu icon 
* added possibility to set Bookmarks Toolbar position
* little changed all-new settings layout (added columned settings, etc.)
* moved WebRTC and HTTP Referrer Header to better place

laniakea64:
Something in this pull request has resulted in tabs not filling up the whole width of the tab bar, leaving empty space on the end.  Setting "Window Controls Position" to "Left side" moves this empty space to the left of all the open tabs.

MrAlex94:
Thanks for the report. Will see what I can reproduce.

Issue #1021, opened 2019-06-11
Waterfox and website lacks obligatory legal information


ilu33:
Since I looked for a way to contact @MrAlex94 I realized that the website lacks legal information about Waterfox Ltd: Address, legal representative, contact information. This needs to be included in the "legal" and "privacy" page.
This missing information could cause legal trouble. And there's no reason to not disclose it anyway because GB has a public company register - just without email contact.

Also the Waterfox page about:buildconfig needs to link to the sources needed for the build as required by MPL 3.2 - see FAQ7 at https://www.mozilla.org/en-US/MPL/2.0/FAQ/. I could not test this on Waterfox 68. Of course you can choose another place (I could not find any) to link the source but that would not make much sense.

grahamperrin:
Meta/tracking: #582

PR #1292, opened 2019-12-06
[current] Include sourcestamp info in builds

laniakea64:
Tag https://github.com/MrAlex94/Waterfox/issues/1021 .  Also, this is very useful for troubleshooting regressions and for people who self build Waterfox.


MrAlex94:
Wicked, thank you.

PR #1303, opened 2019-12-12
(WF Current) Various performance tweaks.

Peacock365:
Enable WebRender, enable OpenGL Off-Main-Thread Compositing (OMTC), enable accelerated Azure Canvas.

Much better performance when these three are enabled. The effect of accelerated Azure Canvas can be seen here: http://js1k.com/2016-elemental/demo/2445

Waterfox is known as a high performance browser; we should enable beneficial performance settings if possible.


laniakea64:
What hardware/OS combinations have you tested this on?

`layers.acceleration.force-enabled` quietly overrides user's hardware acceleration setting in `about:preferences`, which will result in confusion if user didn't flip this pref themselves.  And having that hardware acceleration enabled has caused graphics glitches on my computer.  I think it would be unwise to enable this by default for everyone.

Issue #515, opened 2018-04-09
Silently-installed Mozilla addons


chocolateboy:
Just noticed in Waterfox 56.1.0/Linux (Arch) that the following addons have been silently installed:

* [ADB Helper](https://github.com/mozilla/adbhelper)
* [Valence](https://github.com/mozilla/valence)

(They may have been installed in an earlier version. The timestamp on the `~/.waterfox/<profile>/extensions/fxdevtools-adapters@mozilla.org/` folder is from February, when I was using 56.0.4.1.)

There's a discussion about them [here](http://techdows.com/2015/08/shocking-firefox-silently-installs-adb-helper-and-valence-add-ons.html) (I don't remember ever using WebIDE) and a Bugzilla issue [here](https://bugzilla.mozilla.org/show_bug.cgi?id=1114380).

They look like they could both be useful (or [were](https://bugzilla.mozilla.org/show_bug.cgi?id=1314811)), but I'd rather not have addons installed without my consent. If it's really the best option, it feels like the user should, at the very least, be asked or notified, and that this behaviour should be restricted to nightlies and/or developer builds rather than release builds like Waterfox.

chocolateboy:
According to [this post](https://news.ycombinator.com/item?id=10840289) (which is also discussed on the [Pale Moon forum](https://forum.palemoon.org/viewtopic.php?t=18681)), these addons can be (partially?) disabled by setting the following preferences:

```javascript
pref("devtools.webide.enabled", false)
pref("devtools.webide.autoinstallADBHelper", false)
pref("devtools.webide.autoinstallFxdtAdapters", false)
pref("devtools.webide.autoConnectRuntime", false)
```

MrAlex94:
Also I’ve pointed the pref that checks for system add-ons to download to a
blank XML file. Any chance the Linux distro might be force installing them,
like Ubuntu does?

Otherwise I’ll check if they’re on other distros/OS.

chocolateboy:
Thanks!

> Any chance the Linux distro might be force installing them, like Ubuntu does?

I don't think so. It looks like it's an [unmodified build](https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=waterfox-bin&id=e062bd5d34f5db374189828d84171005372b40d0).

MrAlex94:
I've had a look and can't find these installed on any of my macOS, Windows or Debian systems

chocolateboy:
[...]

Issue #1167, opened 2019-09-24
Waterfox Current help menu


grahamperrin:
Meta, tracking: #538 

### Waterfox help

`about:preferencesfirefox-help`

❗ Invalid URL

(Compare with #400 for Waterfox Classic.)

### Waterfox tour

This menu item is non-effective. 

### Keyboard shortcuts

`about:preferenceskeyboard-shortcuts`

❗ Invalid URL

### Submit feedback…

`#`

❗ Invalid URL

(Compare with #579 for Waterfox Classic.)

### Report deceptive site…

https://en-us.phish-report.mozilla.com/ is untrusted, this is a Mozilla bug.

PR #1304, opened 2019-12-12
[current] misc minor cleanup

laniakea64:
**Disable building leftover shield component**

Waterfox already hard-disables / doesn't have Shield.  This also gets rid of the dead `about:studies` page and unused Shield-related code from Waterfox builds.

**Disable silent auto-install of WebIDE-related extension**

Tag https://github.com/MrAlex94/Waterfox/issues/515 .  In addition to what was said there, Waterfox Current also offers a button in `about:debugging` to install this addon ("Enable USB Devices").  There's no need for it to be _silently_ installed.

**Get rid of some moot about:config listings**

The code that uses these prefs isn't even being called at all in Waterfox.  They are just clutter in `about:config`, making users think something is there to be configured when it's actually not.

Additionally, in the case of `extensions.screenshots.upload-disabled`, Firefox disabled the upload service in Firefox 67 and the upload service was shut down anyway.

**Remove "Pages saved to Pocket" preference UI and default the pref to false**
**Don't show preferences UI for hard-disabled "Recommended by Pocket"**

These two commits get rid of dead Pocket-related preferences UI from `about:preferences#home`.

### DIFF
--- a/browser/app/profile/firefox.js
+++ b/browser/app/profile/firefox.js
@@ -212,11 +212,8 @@ pref("browser.eme.ui.enabled", false);
 // UI tour experience.
 pref("browser.uitour.enabled", false, locked);
 pref("browser.uitour.loglevel", "Error", locked);
-pref("browser.uitour.requireSecure", false, locked);
 pref("browser.uitour.themeOrigin", "", locked);
 pref("browser.uitour.url", "", locked);
-// How long to show a Hearbeat survey (two hours, in seconds)
-pref("browser.uitour.surveyDuration", 7200);
 
 pref("keyword.enabled", true);
 pref("browser.fixup.domainwhitelist.localhost", true);
@@ -1817,9 +1814,6 @@ pref("browser.suppress_first_window_animation", true);
 
 // Preference that allows individual users to disable Screenshots.
 pref("extensions.screenshots.disabled", false);
-// Preference that allows individual users to leave Screenshots enabled, but
-// disable uploading to the server.
-pref("extensions.screenshots.upload-disabled", false);
 
 // URL for Learn More link for browser error logging in preferences
 pref("browser.chrome.errorReporter.infoURL",
@@ -1831,7 +1825,6 @@ pref("app.normandy.dev_mode", false, locked);
 pref("app.normandy.enabled", false, locked);
 pref("app.normandy.first_run", false, locked);
 pref("app.normandy.logging.level", 50); // Warn
-pref("app.normandy.run_interval_seconds", 21600); // 6 hours
 pref("app.normandy.shieldLearnMoreUrl", "", locked);
 #ifdef MOZ_DATA_REPORTING
 pref("app.shield.optoutstudies.enabled", false, locked);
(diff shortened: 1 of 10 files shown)


Peacock365:
There are also other things that need to be cleaned up, maybe you could have a look at them @laniakea64, as I am not knowledgeable enough to remove them:

about:preferences#home still has a "Pages saved to Pocket" entry, the related pref is _browser.newtabpage.activity-stream.section.highlights.includePocket_ and that I would lock to _false_ anyway. The UI entry should go away, too, though.

_security.family_safety.mode_ must be set to 0, in order to fend off potential issues due to local Man in the middle setups. See the Pale Moon 28.4.1 release notes, too: https://www.palemoon.org/releasenotes-archived.shtml Waterfox 56 had this set to 0 as well, as opposed to Firefox where this has been set to 2 since Firefox 52.

laniakea64:
> about:preferences#home still has a "Pages saved to Pocket" entry, the related pref is _browser.newtabpage.activity-stream.section.highlights.includePocket_ and that I would lock to _false_ anyway. The UI entry should go away, too, though.

:+1: 
I was able to remove this UI entry, but so far haven't found where the pref is being set to add the lock.  I can add this UI cleanup to this PR after I've done more testing.  I'll keep looking into the pref, but that might need to wait for a follow-up.

note to self: also look into removing the half-disabled "Recommended by Pocket" pref UI on the same screen.

> _security.family_safety.mode_ must be set to 0, in order to fend off potential issues due to local Man in the middle setups. See the Pale Moon 28.4.1 release notes, too: https://www.palemoon.org/releasenotes-archived.shtml Waterfox 56 had this set to 0 as well, as opposed to Firefox where this has been set to 2 since Firefox 52.

I think this should be handled in a separate issue or PR, as this isn't just "cleanup".  If I get the chance I would try to look into this later.

Edit: I did try to look into this, but it's too Windows-specific for me to actually help with (I only have Linux).

Peacock365:
Good to know that you are looking into them, also nice work so far. I've also noticed that Waterfox Current still contains "Submit Feedback" and "Waterfox Tour" entries in its "Help" menu, both in the menu bar and  in the hamburger menu. Both were removed in Waterfox 56, here is the related thread which points to the commits that were applied to Waterfox 56 in order to remove them:

https://github.com/MrAlex94/Waterfox/issues/548#issuecomment-391214627

laniakea64:
Unfortunately after what I've done so far (some of which still needs more testing before being pushed) I don't have enough time or energy to deal with any other stuff as part of this pull request :frowning_face: Looks like the Help menu could use quite a bit of cleanup, I see https://github.com/MrAlex94/Waterfox/issues/1167 was already filed for that.

> the related pref is _browser.newtabpage.activity-stream.section.highlights.includePocket_ and that I would lock to _false_ anyway. 

I did find where this pref is set, it's not done in the "normal" way so I'm not completely sure the best way to lock it, but I was at least able to change the default to `false`.

laniakea64:
That should do get rid of dead Pocket-related UI from `about:preferences`.

Peacock365:
Very nice work so far, @laniakea64! Things are developing in the right direction. As for the "Help" menu, I found this commit by Toromino:

https://github.com/Toromino/Waterfox/commit/1f8aa8009fe65398cb0e0c6c1b31ecd5e0871953

This got rid of the "Submit Feedback" entry in Waterfox 56, no idea whether or not it still works.

Issue #204, opened 2017-09-03
Privacy Patch: Don't expose navigator.AddonManager to content (from tor browser)


ilikenwf:
While other Tor privacy patches and commits exist, this one is yet another that will prevent "privileged" sites from getting a list of our installed addons. This one doesn't seem to be slated for upstream use in Firefox mainline yet.

There are others like this one that would be beneficial since most are used with options, IE the one that allows setting the max number of fonts used per page.

https://gitweb.torproject.org/tor-browser.git/patch/?id=5493716

> With https://bugzilla.mozilla.org/show_bug.cgi?id=1245571 support for
websites to learn about installed add-ons landed. Currently, this is
only enabled for AMO related sites but we don't think this functionality
is something we want for Tor Browser as it might aid in fingerprinting
users.
>
>The patch does not outright disable access to the API. Privileged code
is still able to use it if needed. This should help with usability
issues should they arise while mitigating possible fingerprinting and
security problems by having this API available to content.

MrAlex94:
Will look into this as well. Will probably remove it as we move away from the AMO.

ilikenwf:
Nice - so you're going to start hosting the "full" addons eventually?

Either way there are lots of tor browser commits we could probably fit into
Waterfox as many of them just harden privacy and security without breaking
any functionality.

ilikenwf:
@MrAlex94 in v68 we should default privacy.resistFingerprinting.block_mozAddonManager to true as they added a hidden pref for it.

This and many other things are already set in the ghacks user.js file - it would be nice to see some kind of integration of that project and the user-overrides method someday...

Until then a few of my tickets could probably be solved using settings from ghacks user.js

laniakea64:
> in v68 we should default privacy.resistFingerprinting.block_mozAddonManager to true as they added a hidden pref for it.

Just to point out, Waterfox 56 has this pref too - https://github.com/MrAlex94/Waterfox/pull/449

ilikenwf:
I guess we can leave it open so we can default both.

laniakea64:
> in v68 we should default privacy.resistFingerprinting.block_mozAddonManager to true 

Unfortunately it looks like making this default isn't viable.  Apparently that broke AMO for some Mac OS users - https://github.com/MrAlex94/Waterfox/issues/1350

PR #449, opened 2018-02-21
privacy.resistFingerprinting.block_mozAddonManager

laniakea64:
Fixes https://github.com/MrAlex94/Waterfox/issues/441

PR #1305, opened 2019-12-12
[current] Expose a couple useful hidden prefs in about:config by default

### LINKED ISSUE #1081: v68: Expose privacy.firstparty.isolate.block_post_message in about:config by default
https://github.com/MrAlex94/Waterfox/blob/f8a37ef0e898f9199f050559ac6bdf931b35a93d/caps/OriginAttributes.h#L95-L99

https://github.com/MrAlex94/Waterfox/blob/f8a37ef0e898f9199f050559ac6bdf931b35a93d/caps/OriginAttributes.cpp#L34-L36

This boolean pref is not visible in `about:config` by default.  I think it was added in Mozilla bug 1492607, and I don't see any rationale there for why it's completely hidden.  It sounds like it could be an interesting security option, so should Waterfox expose this pref in `about:config` by default?

To be clear, I think Waterfox should leave the default value of this pref as-is.  From this line...
https://github.com/MrAlex94/Waterfox/blob/f8a37ef0e898f9199f050559ac6bdf931b35a93d/caps/OriginAttributes.cpp#L22
... and comments in the Mozilla bug, I believe its default value is `false`, but I'm not 100% sure, so didn't submit a pull request.

laniakea64:
`privacy.firstparty.isolate.block_post_message`
Fixes https://github.com/MrAlex94/Waterfox/issues/1081

`privacy.resistFingerprinting.block_mozAddonManager`
This does also change the default value of this pref as per https://github.com/MrAlex94/Waterfox/issues/204#issuecomment-500084037 .  Not locking this pref since I've _occasionally_ found it useful to temporarily flip it to `false`.

### DIFF
--- a/modules/libpref/init/all.js
+++ b/modules/libpref/init/all.js
@@ -1401,6 +1401,13 @@ pref("privacy.firstparty.isolate",                        false);
 // (top level URLs) can access resources through window.opener.
 // This pref is effective only when "privacy.firstparty.isolate" is true.
 pref("privacy.firstparty.isolate.restrict_opener_access", true);
+// Whether to block postMessage across different first party domains when the targetOrigin is '*'.
+// This pref is effective only when "privacy.firstparty.isolate" is true.
+pref("privacy.firstparty.isolate.block_post_message", false);
+
+// If true, DISABLES navigator.mozAddonManager entirely
+sticky_pref("privacy.resistFingerprinting.block_mozAddonManager", true);
+
 // We automatically decline canvas permission requests if they are not initiated
 // from user input. Just in case that breaks something, we allow the user to revert
 // this behavior with this obscure pref. We do not intend to support this long term.


MrAlex94:
Cool, thank you.

grahamperrin:
### `privacy.resistFingerprinting.block_mozAddonManager`

From <https://www.reddit.com/r/waterfox/comments/bt4nm7/increasing_the_adaptability_of_addonsmozillaorg/ese7s24/>

> ### Downside
> 
> If the preference is `true` then you lose the ability to tell whether an extension is installed …

From <https://github.com/MrAlex94/Waterfox/issues/204#issuecomment-500088692>:

> … Just to point out, Waterfox 56 has this pref too - #449

Never seen by me on FreeBSD-CURRENT. 

Build `20191210201058`:

image

What's true for Waterfox Classic on FreeBSD-CURRENT is true for Firefox 71.0 on FreeBSD-CURRENT.

If the preference then `true`, then AMO loses functionality: 

image

When `false` (or not set): 

image

laniakea64:
> > ### Downside
> > If the preference is `true` then you lose the ability to tell whether an extension is installed …

@grahamperrin Yep, that's exactly why I've "_occasionally_ found it useful to temporarily flip it to `false`."

Speaking as someone who has made this a conscious choice, and at times chosen in favor of the AMO feature, I figure the emphasis on "occasionally" made clear enough that the downsides of setting the pref to `true` don't make AMO less usable.  (Especially since the option remains available.)

> > … Just to point out, Waterfox 56 has this pref too - #449
> 
> Never seen by me on FreeBSD-CURRENT.
> ...
> What's true for Waterfox Classic on FreeBSD-CURRENT is true for Firefox 71.0 on FreeBSD-CURRENT.
> ...

:confused: Sounds like what you're seeing is that Waterfox Classic does have this pref, Waterfox Classic just doesn't expose it in `about:config` by default.

grahamperrin:
> Waterfox Classic just doesn't expose it in `about:config` by default.

Ah, thanks for the explanation.

PR #1320, opened 2019-12-22
[Classic] Fix build with Rust 1.40.

### LINKED ISSUE #1315: Waterfox-classic fails to build with rust-1.40
```
   Compiling encoding_glue v0.1.0 (/var/tmp/portage/www-client/waterfox-2019.12/work/Waterfox-2019.12-classic/intl/encoding_glue)
error[E0713]: borrow may still be in use when destructor runs
   --> /var/tmp/portage/www-client/waterfox-2019.12/work/Waterfox-2019.12-classic/third_party/rust/url/src/form_urlencoded.rs:248:40
    |
246 | impl<'a> Target for ::UrlQuery<'a> {
    |      -- lifetime `'a` defined here
247 |     fn as_mut_string(&mut self) -> &mut String { &mut self.url.serialization }
248 |     fn finish(self) -> &'a mut ::Url { self.url }
    |                                        ^^^^^^^^ - here, drop of `self` needs exclusive access to `*self.url`, because the type `UrlQuery<'_>` implements the `Drop` trait
    |                                        |
    |                                        returning this value requires that `*self.url` is borrowed for `'a`

error: aborting due to previous error

For more information about this error, try `rustc --explain E0713`.
error: could not compile `url`.
warning: build failed, waiting for other jobs to finish...
error: build failed
make[4]: *** [/var/tmp/portage/www-client/waterfox-2019.12/work/Waterfox-2019.12-classic/config/rules.mk:998: force-cargo-library-build] Error 101
```

hawkeye116477:
Fixes #1315


Polynomial-C:
:+1: 

Works. Thank you :smiley:

hawkeye116477:
My pleasure.
I needed to use Rust 1.36 to vendor dependencies to fix problem with Rust 1.40... So seems that's time to fix using native `cargo vendor` to avoid switching to Rust 1.36 when will be needed to fix similar problem in the future again and also migrate to python3 :smile:

Polynomial-C:
Good luck with that python3 challenge... I'm quite sure that will become a tough one.

PR #1349, opened 2020-01-10
[Classic] Bug 1381197

### LINKED ISSUE #1346: [Classic] First party isolate breaks cookie cleaner extensions
**Describe the bug**

First party isolation makes self destructing cookies or cookie autodelete unable to clear cookies.

**To Reproduce**
1. privacy.firstparty.isolate to true
2. try clearing cookies
3. cookies "cleared"
4. Check preferences, cookies remain

**Expected behavior**
Isolated cookies are all cleared.

**My configuration** 
 - Device: PC
 - OS: Win7/Linux
 - Version: Classic 2019.12/56

**Additional context**
I run current on some systems and I don't think the bug is present there but will now double check.

hawkeye116477:
This fixes #1346

### DIFF
--- a/toolkit/components/extensions/ext-cookies.js
+++ b/toolkit/components/extensions/ext-cookies.js
@@ -10,6 +10,10 @@ XPCOMUtils.defineLazyModuleGetter(this, "Services",
 
 /* globals DEFAULT_STORE, PRIVATE_STORE */
 
+var {
+  ExtensionError,
+} = ExtensionUtils;
+
 const convertCookie = ({cookie, isPrivate}) => {
   let result = {
     name: cookie.name,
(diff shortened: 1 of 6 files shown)


Peacock365:
Off-topic:

@hawkeye116477, would it be possible to add the option to disable updates of the browser entirely (was removed in Firefox 63):

https://www.ghacks.net/2018/07/28/mozilla-makes-it-more-difficult-to-block-firefox-updates/

Having Live Bookmarks and RSS feed reader back would also be great, if even possible:

https://www.ghacks.net/2018/07/25/mozilla-plans-to-remove-rss-feed-reader-and-live-bookmarks-support-from-firefox/

PR #1353, opened 2020-01-12
[Current] Fix problems related with Tab Bar and Window Controls.

hawkeye116477:
This removes unnecessary free space and also fixes visibility of window controls when they are in left side when menubar enabled. These fixes were only tested on Linux.

### DIFF
--- a/browser/app/profile/firefox.js
+++ b/browser/app/profile/firefox.js
@@ -1923,7 +1923,9 @@ pref("browser.preferences.useAllNew", false);
 pref("extensions.getAddons.showPane", false);
 
 // Window controls position
+#ifdef XP_LINUX
 pref("browser.windowControls.position", "right");
+#endif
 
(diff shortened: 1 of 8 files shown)


laniakea64:
I'm testing this PR and it's much better - I confirm this fixes https://github.com/MrAlex94/Waterfox/pull/1289#issuecomment-571384643 .  Thanks @hawkeye116477 :+1: 

> also fixes visibility of window controls when they are in left side when menubar enabled.

Regarding https://github.com/MrAlex94/Waterfox/pull/1353/commits/d2bae6a653e7d3ee23658a8177dd4d0f0794d869#diff-8239c4c8b7a9599dc99e8834975a42e5R598-R600 -
For me, if I enable `browser.tabs.drawInTitlebar`, seems to need 100px `padding-left` there to fully remove the overlap with titlebar buttons.

PR #1371, opened 2020-01-23
[Current] Make layout simpler and more faithful to original.

hawkeye116477:
Now should be fine on all OS.
But seems that's little bald on top when tab bar is not on default position, maybe should be title.


MrAlex94:
Works, thank you!

PR #1375, opened 2020-01-26
[Current] Fix links to help.

hawkeye116477:
This fixes for example links in addon manager to information about private mode.

### DIFF
--- a/browser/app/profile/firefox.js
+++ b/browser/app/profile/firefox.js
@@ -974,7 +974,7 @@ pref("toolkit.crashreporter.infoURL",
      "", locked);
 
 // base URL for web-based support pages
-pref("app.support.baseURL", "about:preferences");
+pref("app.support.baseURL", "https://support.mozilla.org/1/firefox/%VERSION%/%OS%/%LOCALE%/");
 
 // base url for web-based feedback pages
 #ifdef MOZ_DEV_EDITION


laniakea64:
It's not that simple.

Aside from not knowing why @MrAlex94 set this pref to `about:preferences` in the first place:

1) In Firefox this pref value is `https://support.mozilla.org/1/firefox/%VERSION%/%OS%/%LOCALE%/` (68.0 version is not hard-coded).  Sure, it gives the same result now, but:
 - Waterfox Current isn't always going to be based on Firefox 68.
 - Firefox fills in the full app version there, not just major version.  And Waterfox Current is not based on the outdated "68.0", it's based on some recent version of 68 ESR.  Likely the reason Waterfox is saying "68.0" is because the internal versions in `browser/config/version.txt` and `config/milestone.txt` are out of sync with upstream.  I don't know if this discrepancy would also cause some help pages to show inapplicable information or incorrectly flag Waterfox Current as an outdated Firefox.

2) This causes Help > **Waterfox Help** to go to Firefox help site, which doesn't seem right.

hawkeye116477:
`Firefox fills in the full app version there, not just major version. And Waterfox Current is not based on the outdated "68.0", it's based on some recent version of 68 ESR. Likely the reason Waterfox is saying "68.0" is because the internal versions in browser/config/version.txt and config/milestone.txt are out of sync with upstream. I don't know if this discrepancy would also cause some help pages to show inapplicable information or incorrectly flag Waterfox Current as an outdated Firefox.`

Rather no change, for ESR usually are only security fixes.

`This causes Help > Waterfox Help to go to Firefox help site, which doesn't seem right.`
Currently there is no help page for Waterfox Current. In case of Classic, there is similar situation.
There is forum, but it's not the same, maybe some tips mentioned on Mozilla site can be useful for other people. Even Moonchild left link to Mozilla's page.


`Waterfox Current isn't always going to be based on Firefox 68.`
On this year, I rather won't expect that will be based on some newer version. Even if new version will come then it's possible that could be made again from scratch. Anyway, fine, I changed that.

MrAlex94:
I've added this, so at least there are *some* helpful docs for users. Ideally in the future, will have own support docs set up.

Issue #944, opened 2019-05-07
Unable to build gecko68 branch


laniakea64:
Attempting to build from https://github.com/MrAlex94/Waterfox/commit/af8d60305611159dea4ca8fbf36fa532151f8c5d (gecko68 branch) fails with this error -
```
toolkit/library/libxul.so
../../dom/media/ipc/RDDChild.o: file not recognized: File format not recognized
clang: error: linker command failed with exit code 1 (use -v to see invocation)
/media/gecko/waterfox/Waterfox/config/rules.mk:667: recipe for target 'libxul.so' failed
make[5]: *** [libxul.so] Error 1
/media/gecko/waterfox/Waterfox/config/recurse.mk:74: recipe for target 'toolkit/library/target' failed
make[4]: *** [toolkit/library/target] Error 2
/media/gecko/waterfox/Waterfox/config/recurse.mk:32: recipe for target 'compile' failed
make[3]: *** [compile] Error 2
/media/gecko/waterfox/Waterfox/config/rules.mk:400: recipe for target 'default' failed
make[2]: *** [default] Error 2
client.mk:125: recipe for target 'build' failed
make[1]: *** [build] Error 2
```

Building on Xubuntu 18.04 64-bit, with clang 8.0.1 + nodejs 12.1.0 + rust 1.34.1

MrAlex94:
Remove --enable-lto and it should fix it.

May need to remove this for Linux, or switch the linker.

laniakea64:
Success!  Adding `ac_add_options --disable-lto` to my mozconfig customisations got it to build.  Thanks!

hawkeye116477:
@laniakea64 @MrAlex94 LTO needs more open files, just do `ulimit -n 4096` and it will work :smile:.

laniakea64:
> LTO needs more open files, just do `ulimit -n 4096` and it will work :smile:.

@hawkeye116477 Where would I put that code?  Is it to be added to mozconfig (and if so how?), or is it a shell command to run before starting the build?

hawkeye116477:
@laniakea64 It's a shell command to run before starting the build.

laniakea64:
@hawkeye116477 Thanks for your help, I was able to successfully build leaving `--enable-lto`! :smiley: 

The shell command alone wasn't enough.  Build spewed more errors than I've ever seen anything spew.  So I looked at hawkeye116477's mozconfig, based on that I tried adding these lines to my mozconfig customisations -
```
[...]

PR #1378, opened 2020-01-27
setup dockerfile for linux build environment

fdavis:
grabbed more deps from bootstrap

but problbably will just use bootstrap instead

bootstrap!

one env line per build... only want one build atm

### DIFF
--- a/.mozconfig
+++ b/.mozconfig
@@ -57,11 +57,11 @@ ac_add_options --disable-verify-mar
 ac_add_options --enable-application=browser
 #Crashes PPC64LE builds
 if test `uname -m` != ppc64le; then
-ac_add_options --enable-ccache=sccache
+ac_add_options --enable-ccache=ccache
 elif test `uname -m` = ppc64le; then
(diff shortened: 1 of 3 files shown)


.mozconfig
```diff
@@ -57,11 +57,11 @@ ac_add_options --disable-verify-mar
 ac_add_options --enable-application=browser
 #Crashes PPC64LE builds
 if test `uname -m` != ppc64le; then
-ac_add_options --enable-ccache=sccache
+ac_add_options --enable-ccache=ccache
 elif test `uname -m` = ppc64le; then
 ac_add_options --enable-ccache=ccache
 fi
-ac_add_options --enable-lto
+#ac_add_options --enable-lto
```
laniakea64, line 64: Is this the way the official Waterfox Current Linux builds are avoiding https://github.com/MrAlex94/Waterfox/issues/944 ?

MrAlex94: Yes that’s right - I’ve got a local mozconfig that’s slightly different

PR #1381, opened 2020-01-28
(Current) Privacy: Don't reveal internal IP address when WebRTC is en…

Peacock365:
…abled.

See here:

 https://github.com/beefproject/beef/wiki/Module%3A-Get-Internal-IP-WebRTC


MrAlex94:
This will break Google WebRTC products. Not sure it's worth turning off by default.

Peacock365:
I see. Closing this pull request as I realize that it might lead to too much breakage. We already have a visible setting to turn off WebRTC completely in the Waterfox preferences, this should suffice for people who do not need WebRTC anyway.

ognone:
@MrAlex94 Would it be possible to compile Waterfox without WebRTC? Not only disabled in prefs, but completely patched out?

PR #1382, opened 2020-01-28
(Current) Privacy: Delete temporary files on exit (macOS).

Peacock365:
Align the behavior of Waterfox under macOS with that under Windows and Linux. See here:

https://bugzilla.mozilla.org/show_bug.cgi?id=238789


MrAlex94:
Not sure I understand what this preference does. Could you give an example? I can see how it happens for Thunderbird but am unaware for Waterfox.

PR #1385, opened 2020-01-31
fixup docker environment

fdavis:
dockerfile.linux image can now be used to build
current or classic for linux

nospam in the build env

path update to find rust

classic missing gconf2.0

run rustup silent

bugfix rust no pront

use rust 1.37.0 in builds

ordering bugfix, rustup ubuntu


hawkeye116477:
As I know, gconf is deprecated, even Ubuntu Mozilla Team disables it in config.

PR #1403, opened 2020-02-13
[Current] Bug 1587922 - Remove the remnants of the showModalDialog code.

hawkeye116477:
Similarly as for Classic :smile:.

#wintercleaning


MrAlex94:
🧹

PR #1440, opened 2020-02-18
(Classic) Privacy: Get rid of SSL error telemetry.

Peacock365:
See header.


hawkeye116477:
Seems that this breaks Waterfox Classic. I got `memory protection violation`.

PR #1445, opened 2020-02-19
[Classic] Fix memory protection violation.

hawkeye116477:
@Peacock365 Not cool. 

I didn't checked Current, but maybe it will need same fix.


Peacock365:
Any idea whether or not Current also needs this fix?

hawkeye116477:
@Peacock365 Seems that it doesn't need that fix.

PR #1426, opened 2020-02-16
[Classic] Update Async Iteration to follow the proposal change and let it ride the train.

hawkeye116477:
This fixes #1424.

:train2: 

Small update: Additionally added few performance tweaks and removed unboxed arrays dead code.

PR #1447, opened 2020-02-19
Revert "[Classic] Update Async Iteration to follow the proposal change and let it ride the train."

MrAlex94:
Reverts MrAlex94/Waterfox#1426


hawkeye116477:
What's the problem?

PR #1460, opened 2020-02-29
(Classic) Privacy: Disable link prefetching, disable DNS prefetching,…

Peacock365:
… disable predictor.

See here ("Prefetching" section):
https://support.mozilla.org/en-US/kb/how-stop-firefox-making-automatic-connections

And here:
https://www.ghacks.net/2013/04/27/firefox-prefetching-what-you-need-to-know/

Improves privacy by reducing speculative connections Waterfox might otherwise establish. Supposedly comes with a performance penalty, though I haven't noticed anything in this regard even on script-heavy websites. No website breakage whatsoever is to be expected here.

### DIFF
--- a/modules/libpref/init/all.js
+++ b/modules/libpref/init/all.js
@@ -2023,7 +2023,7 @@ pref("network.dnsCacheExpirationGracePeriod", 60);
 pref("network.dns.disablePrefetch", true);
 
 // This preference can be used to turn off DNS prefetch (HTTPS).
-pref("network.dns.disablePrefetchFromHTTPS", true)
+pref("network.dns.disablePrefetchFromHTTPS", true);
 
 // This preference controls whether .onion hostnames are
 // rejected before being given to DNS. RFC 7686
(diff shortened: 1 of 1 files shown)


Squall-Leonhart:
@MrAlex94  please stop adding changes from the tinfoil hat brigade, this change has made mangareaders and general samesite browsing slower as they no longer prefetch the next page,  put it under an  enhanced privacy pref or something but do not make it the default.

Peacock365:
@Squall-Leonhart 

This change clearly benefits privacy and all privacy guides for Firefox recommend changing it. Sad to say, but when I read the name @Squall-Leonhart in my inbox, I already know that the related comment of yours will not be worth reading and a total waste of time. Your presence here doesn't benefit this project, I think of you as a troll already.

Squall-Leonhart:
Waterfox's core tenant is not privacy, you tinfoil hattists have hijacked development and are degrading the experience.

Your presence on this earth is just a troll to the planet.

Peacock365:
Waterfox is being advertised as privacy friendly, @Squall-Leonhart. If disabling link prefetching and DNS prefetching "degrades" your experience, feel free to turn them back on. But then, if the minimal impact of prefetching already "degrades" your experience, then your connection must be extremely lousy / bad. With above ISDN speed, the impact should be negligible.

PR #1463, opened 2020-02-29
(Classic) Privacy: Enable tracking protection by default.

Peacock365:
See header. Enables tracking protection, only blocks known tracking cookies. No website breakage is expected here, since Waterfox by default uses the normal blocklist instead of the strict one.

### DIFF
--- a/modules/libpref/init/all.js
+++ b/modules/libpref/init/all.js
@@ -1389,7 +1389,7 @@ pref("privacy.donottrackheader.enabled",    false);
 // the popup will be called when the popup is dismissed.
 pref("privacy.permissionPrompts.showCloseButton", false);
 // Enforce tracking protection in all modes
-pref("privacy.trackingprotection.enabled",  false);
+pref("privacy.trackingprotection.enabled",  true);
 // Enforce tracking protection in Private Browsing mode
 pref("privacy.trackingprotection.pbmode.enabled",  true);
 // Annotate channels based on the tracking protection list in all modes


hawkeye116477:
I'm not sure if this good idea. It could interfere with adblockers and I didn't see a way to disable it from preferences. So if you want to change that, for first, switch option should be added to preferences.

`No website breakage is expected here, since Waterfox by default uses the normal blocklist instead of the strict one.`

Rather it still can break websites, I remember when people reported problems, because of that.

grahamperrin:
> … to disable it from preferences. …

**about:preferences#privacy**

### Breakage

The worst of the edge cases are almost impossible to work around at the site level, it becomes necessary to (temporarily) disable tracking protection completely. 

Right now I can't recall whether those edge cases were encountered in strict mode …

----

… OK, I found [Where, on disk, are tracking exception preferences stored? : firefox](https://old.reddit.com/r/firefox/comments/eeymsf/where_on_disk_are_tracking_exception_preferences/) where I intentionally refrained from mentioning Waterfox (because I wanted a simple, straight answer) but the underlying issue involved Google Translate <strike>misbehaving </strike>in some situation(s) with Waterfox Classic. IIRC: 

* the layouts of translated pages were sometimes quite badly messed up

– and the _Exceptions…_ dialogue at about:preferences#privacy allows removals **but not** addition. 

Unfortunately (sorry) I didn't keep a note of the situations in which mess occurred … from the date, around two months ago, it's likely that I was investigating cybercrime and intentionally not keeping browser histories. If so, it might be impossible for me to make reproducible a _messy layout_ case. 

Given my lack of record-keeping I should probably sit on the fence re: suitability of enabling tracking protection by default. Excellent idea, though. 

### Conflicts

> … could interfere with adblockers …

FWIW: to the best of my knowledge, never a problem with Malwarebytes Browser Guard. 

I mean, no sense of interference **however** where there's a multi-layered approach to tracking protection, it can be surprisingly easy to forget the need to perform multiple steps when experimenting to tell whether a protector is (or protectors are) the cause of a problem. 

HTH

hawkeye116477:
Tracking protection is just like another adblocker, having it enabled and some adblocker is rather too much, cuz adblocker already blocks tracking, usually does it even better and tracking protection even on basic level sometimes could break website functionality.

`It is surprisingly easy to forget the need to perform multiple steps when experimenting to tell whether a protector is (or protectors are) the cause of a problem`

That's also argument against this idea.

`Given my lack of record-keeping I should probably sit on the fence re: suitability of enabling tracking protection by default. Excellent idea, though.`

I also can't remember when those reports are, but doesn't mean that I think that's a good idea. 
If already, at least switch on preferences to disable it completely would be handy.

grahamperrin:
> … switch on preferences to disable it completely …

This? Third of the three radio buttons: 

image

(If I misunderstood: sorry, I'll hide this comment.)

hawkeye116477:
@grahamperrin Yes, but I didn't have that.

Update: Seems that there is pref `privacy.trackingprotection.ui.enabled`, which enables showing this.

grahamperrin:
Ah, OK. I have `privacy.trackingprotection.ui.enabled` present and `true` by default. 

Home-built `2019.12` (`20200103060435`) on FreeBSD-CURRENT.

hawkeye116477:
@grahamperrin https://github.com/MrAlex94/Waterfox/blob/classic/browser/app/profile/firefox.js#L1510

grahamperrin:
Yeah, I guess that for the port to FreeBSD, Jan Beich enabled the UI long ago.

Issue #1351, opened 2020-01-12
Window class on Linux makes it impossible to use docks/launchers


Slater91:
After the 2020.01 update, both Waterfox Classic and Waterfox Current identify themselves as "waterfox", so window class-based detection methods (e.g. the one used by Latte Dock) fail.

**To Reproduce**

1. Open Waterfox Current and Waterfox Classic on KDE Plasma
2. Both are identified and treated as Waterfox Classic on Latte Dock
3. Right-click on the titlebar and select "configure special application settings"
4. Both Classic and Current are identified as "waterfox"

**Expected behavior**
The two browsers should identify themselves correctly as "waterfox current" and "waterfox classic" as they did in the previous version.

**My configuration** 
 - OS: Linux - KDE neon
 - Version: Classic + Current 2020.01

Slater91:
Note: it seems like something has changed between the 2019.12 version and the 2020.01 version. By setting "StartupWMClass=WaterfoxClassic" in the .desktop file, the issue goes away. This was not necessary in the previous version, so it seems like some change was made that impacts the ability of WMs to correctly identify the windows.

Issue #1350, opened 2020-01-11
Waterfox 2020.01 is still buggy.


Peacock365:
Using macOS 10.15.2 Catalina and Waterfox 2020.01 here.

MrAlex94:
Does this occur in Safe Mode?

Peacock365:
I have neither add-ons nor themes installed (clean install), so yes.

hawkeye116477:
Probably it's something related with `https://github.com/MrAlex94/Waterfox/commit/6c7310b20bc528c6562d4659c948d052b4fd7f38`, as I see in code some things maybe different in Mac, but I didn't have Mac, so can't test.

MrAlex94:
Hmm, currently on Catalina and can’t reproduce the issues. I’ll comb through and see if I can see what’s affecting older macOS systems.

Peacock365:
So it seems that the close / minimize / maximize button issue on macOS was fixed by: https://github.com/MrAlex94/Waterfox/commit/f7c022c92adff1cee3448068274094bf531601bd

Hopefully the status bar issue I mentioned will also get tackled before a new Waterfox version is being released.

Last but not least, it's still impossible to add extensions from AMO with a clean profile. I do not think that this issue is macOS-specific, so maybe @hawkeye116477 and @laniakea64 could also have a look at it, the efforts of @MrAlex94 notwithstanding?

hawkeye116477:
I had also problems with extensions some day, I clicked to add extension, then it showed error on AMO, but now seems all is fine. 
In case of Status Bar, currently on css it's set to not be visible in full screen mode, just like on Vivaldi.

laniakea64:
> it's still impossible to add extensions from AMO with a clean profile. I do not think that this issue is macOS-specific, so maybe @hawkeye116477 and @laniakea64 could also have a look at it, the efforts of @MrAlex94 notwithstanding?

It looks to work for me in self build from https://github.com/MrAlex94/Waterfox/commit/7ddf207b656e82702eddb3c4c5d2217a9fabaf22 .  What specific extension(s) were you trying to add and what happened instead of them being added?

Peacock365:
> In case of Status Bar, currently on css it's set to not be visible in full screen mode, just like on Vivaldi.

[...]

Issue #1211, opened 2019-10-19
Waterfox Classic using the icon for Current; elsewhere, Waterfox Current using the icon for Classic


grahamperrin:
Bugging builds 20191015002329 (Classic) and 20191015121210 (Current) on KDE Plasma: 

![2019-10-19 09:06:48](https://user-images.githubusercontent.com/192271/67140213-2f840f00-f250-11e9-8877-a25e36af7059.png)

1. Waterfox Classic 2019.10 (20191015002329) in the window to the left
2. Waterfox Current 2019.10 (20191015121210) to the right. 

Also, both applications appear as a single task in Task Manager: 

![2019-10-19 09:07:29](https://user-images.githubusercontent.com/192271/67140260-d7014180-f250-11e9-9330-579a08279857.png)

- Waterfox Classic misusing the icon for Waterfox Current
- Waterfox Classic appearing as _Waterfox_ (not Classic)
- <strike>Waterfox Current apparently nameless</strike> my mistake, I misunderstood the GUI of Task Manager.

Some overlap with: 

- <strike>`#769` (Firefox branding still shown in Waterfox)</strike>
- #1210 (If Waterfox Current 2019.10 runs first, then Waterfox Classic 2019.10 will not run). 

Meta, tracking: #538

hawkeye116477:
I have another situation, Waterfox Current uses Classic icon.

grahamperrin:
LOL poor Alex, poor you! 

What are the build IDs and desktop enviroment in your case?

hawkeye116477:
It's my own build from newest commit and I'm on KDE Plasma.

hawkeye116477:
`misrepresented as Mozilla Firefox in title bars,`
I guess that's after installing langpack from AMO.

grahamperrin:
> I guess that's after installing langpack from AMO.

Exactly! Now I understand, thanks. https://github.com/MrAlex94/Waterfox/pull/1213#issuecomment-544143709 down-voting myself. 

A screen recording – beginning with a preference for English (United Kingdom), switching to English (United States): 

- [2019-10-19 19:01.zip](https://github.com/MrAlex94/Waterfox/files/3747035/2019-10-19.19.01.zip)

aquaspy:
Any fix to this? :(

grahamperrin:
@aquaspy does Waterfox Classic 2019.12 have the issue?

grahamperrin:
[...]

PR #1466, opened 2020-02-29
[Current] Few fixes

hawkeye116477:
Fixed:
* icon mish mash on Linux (#1211, #1351)
* Widevine crashes with glibc 2.31
* https://github.com/MrAlex94/Waterfox/issues/1350#issuecomment-592927446

Related => https://bugzilla.mozilla.org/show_bug.cgi?id=1530052

Updated mozonfig also (most people rather use packages from repositories, not mozbuild)

### DIFF
--- a/.mozconfig
+++ b/.mozconfig
@@ -20,9 +20,19 @@ export WIN_UCRT_REDIST_DIR="/c/Program Files (x86)/Windows Kits/10/Redist/ucrt/D
 elif test `uname -s` = Linux; then
 #x86 Specific Options
 if test `uname -m` = x86_64; then
+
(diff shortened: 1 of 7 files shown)


grahamperrin:
Overlap with <https://old.reddit.com/r/waterfox/comments/fb7oe4/drm_aint_working/fj2zlri/>? 

> … widevine appears to be installed but Spotify still refuses to play and …

Or separate? 

In any case: thanks 👍

hawkeye116477:
I see now that Debian 10 which is used by TransistorSnack has 2.28 and that problem I mentioned appears on 2.31, so it's rather different problem.

PR #1468, opened 2020-03-01
(Classic) Enable playback of encrypted VP9 video in fMP4 container

Peacock365:
Lock some telemetry and experiments and crashreporter settings as well.


hawkeye116477:
@Peacock365 I hope that you checked now that locking doesn't causes build errors :smile:.

PR #1484, opened 2020-03-09
(Current) Clean up the prefs a bit.

Peacock365:
See header.


Peacock365:
@MrAlex94 

Is there any good reason for not merging this? I can't seem to find one, it's just a clean-up.

PR #1485, opened 2020-03-09
(Current) Increase performance a bit.

Peacock365:
See header.


MrAlex94:
Not sure on changing the font settings, [see here](https://www.igvita.com/2014/01/31/optimizing-web-font-rendering-performance/#timeouts).

I'll do some more research into the image decoding, but looks like from my provisional research, switching that setting makes the browser consume more memory.

Peacock365:
Well, these settings improve the performance of Waterfox across the board for me.

Peacock365:
Off-topic: When will you fix the CSS of the New Tab button (behind tabs) in your Australis themes? Shouldn't the button be more "rounded"?

PR #1558, opened 2020-05-04
(Current) Privacy: Add privacy.resistFingerprinting.delegateCanvasProtection 

ilikenwf:
Currently Waterfox and Firefox both break CanvasBlocker and other addons that seek to make canvas fingerprinting harder. While people debate over what is best for these situations, recent changes to RFP caused these extensions to no longer fulfill their primary purpose of randomizing the canvas fingerprint hashes.

This tweak adds a default false option of 

```privacy.resistFingerprinting.delegateCanvasProtection``` 

 which when enabled, stops checks of canvas permissions just before Waterfox would normally show or autodecline the request to read the canvas API.

This way, the other prior checks are still enforced, but should the end user, a privacy centric or power user, should desire, they can either disable the rest of the handling of canvas protection entirely (if no addon is configured) or more ideally, use CanvasBlocker to block, randomize, black/whitelist, or display prompts for canvas data extraction where necessary, all while retaining the other benefits that come from using RFP.


ilikenwf:
Closing, I forgot how dirty this branch was, I'll reopen.

PR #1559, opened 2020-05-04
(Current) Privacy: Add privacy.resistFingerprinting.delegateCanvasProtection 

ilikenwf:
Currently Waterfox and Firefox both break CanvasBlocker and other addons that seek to make canvas fingerprinting harder. While people debate over what is best for these situations, recent changes to RFP caused these extensions to no longer fulfill their primary purpose of randomizing the canvas fingerprint hashes.

This tweak adds a default false option of 

```privacy.resistFingerprinting.delegateCanvasProtection``` 

 which when enabled, stops checks of canvas permissions just before Waterfox would normally show or autodecline the request to read the canvas API.

This way, the other prior checks are still enforced, but should the end user, a privacy centric or power user, should desire, they can either disable the rest of the handling of canvas protection entirely (if no addon is configured) or more ideally, use CanvasBlocker to block, randomize, black/whitelist, or display prompts for canvas data extraction where necessary, all while retaining the other benefits that come from using RFP.

If this or something like it were added to Waterfox, I feel many many more users would switch to Waterfox, especially the ghacks user.js crowd and the various privacy communities that are around.

### DIFF
--- a/dom/canvas/CanvasUtils.cpp
+++ b/dom/canvas/CanvasUtils.cpp
@@ -138,6 +138,18 @@ bool IsImageExtractionAllowed(Document* aDocument, JSContext* aCx,
 
   // At this point, permission is unknown
   // (nsIPermissionManager::UNKNOWN_ACTION).
+  
+  // Check if the user has explicitly disabled all
+  // or delegated all other checks and fingerprinting resistence
+  // to an extension (or nothing) such as canvasblocker
+  bool isDelegateCanvasProtection =
+      StaticPrefs::
+          privacy_resistFingerprinting_delegateCanvasProtection() &&
+      !EventStateManager::IsHandlingUserInput();
+
+  if (isDelegateCanvasProtection) {
+    return true;
+  }
 
   // Check if the request is in response to user input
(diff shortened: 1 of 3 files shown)


ilikenwf:
And I'd submit this upstream to Firefox but they seem opposed to any separation here, from what I've read...makes little sense to me since canvasblocker brings the identifiers down quite a bit even compared to stock RFP.

ilikenwf:
Obligatory fingerprinting results with this enabled and canvasblocker randomizing...note this includes a bunch of addons and a custom user.js:

image

image

MrAlex94:
Sorry I haven’t looked at this yet; it’s on the list of things I’ll get round to review before the next release!

MrAlex94:
Appreciate the patch!

ilikenwf:
> Appreciate the patch!

Hey thanks! The schools of thought vary on whether this is helping or not, because of sample bias with the fingerprint test sites, if we ever add documentation for this setting it may be wise to consider Thorin's wise discussion here: https://github.com/ghacksuserjs/ghacks-user.js/issues/942

PR #1583, opened 2020-05-26
[Classic] Add fixes for crashes.

hawkeye116477:
These are few fixes for crashes.

For example previously test from https://github.com/mdn/web-components-examples/blob/master/edit-word/index.html crashed and this site => https://polymer-library.polymer-project.org/1.0/docs/first-element/intro#run-the-demo also crashed, now it's without crashes :smile:


MrAlex94:
Great, thank you!

Issue #1549, opened 2020-04-24
https://broadcast.nvidia.com/feedback?sdk=voice white page


Squall-Leonhart:
White screen in WF Classic 2020.04

also affects
https://broadcast.nvidia.com/feedback?sdk=greenscreen

Reproduced in new profile, not addin related

hawkeye116477:
Just set `webgl.enable-debug-renderer-info` to `true` and it will work.

@Peacock365 Broke it in https://github.com/MrAlex94/Waterfox/pull/1461

Squall-Leonhart:
Thanks, another strike against the collective stupidity of the pro-privacy crowd.

Squall-Leonhart:
fixed

PR #1638, opened 2020-07-01
Port some privacy changes of Waterfox Current to future versions.

Peacock365:
- Disable link prefetching, disable DNS prefetching
- No more information about the graphics driver for debugging purposes.
- Only send pings if send and receiving host match.
- Get rid of SSL error telemetry.
- Disable establishing connection to links the mouse hovers over.
- Enforce Punycode for internationalized domain names.
- Disable some telemetry.

### DIFF
--- a/modules/libpref/init/all.js
+++ b/modules/libpref/init/all.js
@@ -158,9 +158,9 @@ pref("security.webauth.webauthn_enable_softtoken", false);
   pref("security.webauth.webauthn_enable_usbtoken", true);
 #endif
 
-pref("security.ssl.errorReporting.enabled", false);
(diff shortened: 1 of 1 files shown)


modules/libpref/init/all.js
```diff
@@ -3893,7 +3896,7 @@ pref("image.http.accept", "image/webp,*/*");
 // Allows image locking of decoded image data in content processes.
 pref("image.mem.allow_locking_in_content_processes", true);
 
-pref("webgl.enable-debug-renderer-info", true);
+pref("webgl.enable-debug-renderer-info", false);
```
hawkeye116477, line 3899: Maybe better to don't do this, cuz that causes problem #1549.

hawkeye116477 (changes requested):

PR #1745, opened 2020-09-08
[Classic] Set SPOOFED_OSCPU as Windows 10 (like in Current).

aeiouaeiouaeiouaeiouaeiouaeiou:
Cosmetic fix for "resist fingerprinting" mode.

### DIFF
--- a/browser/components/resistfingerprinting/test/browser/browser_navigator.js
+++ b/browser/components/resistfingerprinting/test/browser/browser_navigator.js
@@ -13,7 +13,7 @@ var spoofedUserAgent;
 const SPOOFED_APPNAME        = "Netscape";
 const SPOOFED_APPVERSION     = "5.0 (Windows)";
 const SPOOFED_PLATFORM       = "Win64";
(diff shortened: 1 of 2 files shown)


AM4Ryzen:
Would rather have the spoofed OS be Windows 7. It's better to make it's marketshare "higher" than it is to make Windows 10's marketshare "higher".

Edit: Considering making the opposite of this for Current

aeiouaeiouaeiouaeiouaeiouaeiou:
https://bugzilla.mozilla.org/show_bug.cgi?id=1511434

privacy.resistFingerprinting is not set on by default and only needed in exceptional cases for privacy. I don't think it will affect the marketshare so much.

AM4Ryzen:
> https://bugzilla.mozilla.org/show_bug.cgi?id=1511434
> 
> privacy.resistFingerprinting is not set on by default and only needed in exceptional cases for privacy. I don't think it will affect the marketshare so much.

I want to represent my favorite OS no matter what OS I'm running

PR #1748, opened 2020-09-11
Customise default preferences further.

Peacock365:
- Disable random connections being established by the URL bar.
- Disable URL shortening.
- No longer hide away the download button (as in WF Current).
- Revert to green lock icon (better visibility).
- Disable even more telemetry.
- Disable service workers for privacy reasons.
- Disable scheduled telemetry task.

Some things were probably overseen, the commit is meant to fix those. Comments and corrections appreciated.

### DIFF
--- a/browser/app/profile/firefox.js
+++ b/browser/app/profile/firefox.js
@@ -40,7 +40,7 @@ pref("extensions.getAddons.get.url", "https://services.addons.mozilla.org/api/v3
 pref("extensions.getAddons.search.browseURL", "https://addons.mozilla.org/%LOCALE%/firefox/search?q=%TERMS%&platform=%OS%&appver=%VERSION%");
 pref("extensions.getAddons.link.url", "https://addons.mozilla.org/%LOCALE%/firefox/");
 pref("extensions.getAddons.langpacks.url", "https://services.addons.mozilla.org/api/v3/addons/language-tools/?app=firefox&type=language&appversion=%VERSION%");
-pref("extensions.getAddons.discovery.api_url", "");
+pref("extensions.getAddons.discovery.api_url", "", locked);
 
 // Use bloomfilters for the addons blocklist, instead of JSON only.
 pref("extensions.blocklist.useMLBF", true);
@@ -49,7 +49,7 @@ pref("extensions.blocklist.useMLBF.stashes", true);
 // The URL for the privacy policy related to recommended extensions.
 pref("extensions.recommendations.privacyPolicyUrl", "https://www.mozilla.org/privacy/firefox/?utm_source=firefox-browser&utm_medium=firefox-browser&utm_content=privacy-policy-link#addons");
 // The URL for Firefox Color, recommended on the theme page in about:addons.
-pref("extensions.recommendations.themeRecommendationUrl", "");
+pref("extensions.recommendations.themeRecommendationUrl", "", locked);
 
 pref("extensions.update.autoUpdateDefault", true);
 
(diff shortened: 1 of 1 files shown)


MrAlex94:
Still testing out various config changes in firefox.js (already changed some of what you have), so may just add these changes to my current commit instead of doing a pull request on top.

Peacock365:
Yeah, I'm fine with that. Some further notes:

- I don't recommend locking _browser.send_pings_, because this is website-related and should be possible to toggle. Setting it to false by default is fine, though.
- I recommend disabling service workers by default just like in previous ESRs, for privacy reasons. If somebody wants this, it can still be toggled in about:config. Maybe you could also consider introducing an "Enable Push notifications" toggle in the settings that would set both _dom.serviceWorkers.enabled_ and _dom.push.enabled_ to "true" when it's checked.
- Please also set _services.sync.prefs.sync.browser.discovery.enabled_ to "false / locked", Waterfox aims to completely disable Discovery so this should not be set to true. I forgot about this one in my pull request here.

Allow me a word of explanation: I checked out your project page and Waterfox 78 (new Current) seems to shape up to be a good release, so I decided to give it yet another go and commit to it. All the best for this release.

Peacock365:
Oh and it would also be advisable to get rid of the megabar (various CSS fixes are out there) and to get rid of this scheduled telemetry task entirely:

https://www.ghacks.net/2020/04/09/mozilla-installs-scheduled-telemetry-task-on-windows-with-firefox-75/

One of the settings I've changed makes this telemetry task fail immediately, but not creating it at all would be preferable, creating it in the first place would mean bad optics for Waterfox.

MrAlex94:
> I don't recommend locking browser.send_pings, because this is website-related and should be possible to toggle. Setting it to false by default is fine, though.

Oops, you are correct. I associated that setting with the [Telemetry Ping](docs.telemetry.mozilla.org/datasets/pings.html) setting.

> I recommend disabling service workers by default just like in previous ESRs, for privacy reasons. If somebody wants this, it can still be toggled in about:config. Maybe you could also consider introducing an "Enable Push notifications" toggle in the settings that would set both dom.serviceWorkers.enabled and dom.push.enabled to "true" when it's checked.

I'll see for this one. I may have it on by default, but add it to the on-boarding for the user to disable if they wish.

> Please also set services.sync.prefs.sync.browser.discovery.enabled to "false / locked", Waterfox aims to completely disable Discovery so this should not be set to true. I forgot about this one in my pull request here.

Aye, this has already been toggle off in my working commit.

> Allow me a word of explanation: I checked out your project page and Waterfox 78 (new Current) seems to shape up to be a good release, so I decided to give it yet another go and commit to it. All the best for this release.

Thanks 👍

> Oh and it would also be advisable to get rid of the megabar (various CSS fixes are out there) 

A designer is creating a completely custom UI for Waterfox. Will see what to do in regards to the megabar.

> and to get rid of this scheduled telemetry task entirely:

I think this is disabled compltely, but will verify and make sure it's off.

Peacock365:
@MrAlex94 

_default-browser-agent.enabled_ being set to **false** (like in this commit) only makes the scheduled telemetry task fail immediately. It doesn't prevent that task from being created. Creating it is bad optics for Waterfox though, so if you could prevent its creation, that would be great.

Furthermore, don't forget to thoroughly purge the system add-ons, keeping only AutoFill and the screenshot tool (like in present day Waterfox Current) seems like the sensible choice to me.

MrAlex94:
The browser agent is disabled at build time with `ac_add_options --disable-default-browser-agent`. I've also [set the macro](https://github.com/MrAlex94/Waterfox/commit/12be0d31759319517bfde5d6d7f80e423be82950#diff-af29560da5a3f057628cef675f635140R12) to disable it as well (and now the config option too from your pull request). I've taken some of your changes and added them to this commit https://github.com/MrAlex94/Waterfox/commit/1087e853687b37e5f6447750f21e94284c52e63c, thank you for that 👍. Will close this and your other pull requst.

PR #1761, opened 2020-09-20
Modify default configuration yet again.

Peacock365:
- Excluded Discovery from sync.
- Revert pings being set to true.
- Crash report setting that was forgotten disabled.
- Give users the option to enabled extension signing for security reasons if they so wish.
etc.


MrAlex94:
https://github.com/MrAlex94/Waterfox/commit/6bfcb3654cecbdec057ab44aa69c0512d314dc73

Changed some of what you suggested in this commit.

PR #1762, opened 2020-09-20
Update all.js

Peacock365:
- Disable O-RTT data on TLS 1.3 (as in Classic and Current)
- Get rid of even more prefetching and speculative connections (as in Classic and Current).
- Loosen WebExtensions domain restrictions a bit (makes them work on AMO).
- Disable Web Compatibility reporter feature.
- Various others.


MrAlex94:
6bfcb36

Changed some of what you suggested in this commit. Didn't change everything though, i.e. ability to submit website compatibility issues is useful, and in the future crash reports contain helpful information. I just don't want them uploaded anywhere or automatically either.

PR #1782, opened 2020-10-09
Update firefox.js

Peacock365:
- Revert extension signing being turned off as a hard lock.

Reason: Setting this setting to "false" already achieves what we want (turning off extension signing), users should have the option to re-enable this, if they want to, there are security reasons for extension signing.

- Turn off BITS updater.

Reason: @MrAlex94 has disabled this already, might as well disable the pref for the sake of completeness.

- Turn off auto-hiding of download button.

Reason: Hiding the download button every single time after a download is complete is very bad UI design, as this would turn the download button into an unstable UI element, also confusing for new users who have never used Firefox or Waterfox before. We are not hiding the download button in WF Current, for the sake of consistency and stability, we should not hide it now.

- Bad cert expert enabled.

Reason: Gives more details when users encounter a bad cert, does not limit their options of what to do next as much.

- Plug-ins click to play enabled

Reason: Enabling plug-ins in general would be bad for security reasons, however, disabling them entirely might cause confusion ("I have XYZ plug-in installed, why can't I use it!?"), plug-ins being click to play is a good and workable compromise.

- Send pings reverted

Reason: This is bad for privacy, plain and simple. Not even Firefox ESR does it, so why do we do it? I know that I suggested only allowing pings to the same host, and @MrAlex94 implemented this, which is good, however, I didn't mean to imply that it would be a good ide to enable pings in general. What I wanted to do was the following: Leave them disabled by default, but if users choose to enable them, limit them to the same host to preserve a minimum level of privacy.

- Mobile footer hiding locked to true.

Reason: Since we have removed the correspondent URLs, seeing the footer has no effect for users. Might as well hard-lock this, as we would not want this to get re-enabled, leaving dysfunctional mobile links for the user to look at.

- Various other settings locked to prevent the user from re-enabling them.

NOT SURE ABOUT, STRICTLY NEED THE HELP OF @MrAlex94 FOR THESE:

- browser.uitour.themeOrigin
- browser.uitour.url
- browser.uitour.surveyDuration

Since Waterfox has a custom setup tour now, I am not sure what to do with these. My gut feeling is that it would be a bad idea to leave browser.uitour.surveyDuration as is, however. Do we need this one in particular?

- browser.search.geoSpecificDefaults
- browser.search.geoSpecificDefaults.url

Seems to be an oversight of @MrAlex94 ... Not sure we want to leave those as they are, however, I have no idea what to do with those.

- reader.errors.includeURLs

Is it really necessary to leave this enabled (not sure)?

### DIFF
--- a/browser/app/profile/firefox.js
+++ b/browser/app/profile/firefox.js
@@ -40,16 +40,16 @@ pref("extensions.getAddons.get.url", "https://services.addons.mozilla.org/api/v3
 pref("extensions.getAddons.search.browseURL", "https://addons.mozilla.org/%LOCALE%/firefox/search?q=%TERMS%&platform=%OS%&appver=%VERSION%");
 pref("extensions.getAddons.link.url", "https://addons.mozilla.org/%LOCALE%/firefox/");
 pref("extensions.getAddons.langpacks.url", "https://services.addons.mozilla.org/api/v3/addons/language-tools/?app=firefox&type=language&appversion=%VERSION%");
-pref("extensions.getAddons.discovery.api_url", "");
+pref("extensions.getAddons.discovery.api_url", "", locked);
 
 // Use bloomfilters for the addons blocklist, instead of JSON only.
 pref("extensions.blocklist.useMLBF", true);
 pref("extensions.blocklist.useMLBF.stashes", true);
 
 // The URL for the privacy policy related to recommended extensions.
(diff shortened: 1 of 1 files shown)


MrAlex94:
```
browser.uitour.themeOrigin
browser.uitour.url
browser.uitour.surveyDuration
```

I am unsure whether these are needed yet - but AFAIK they aren't accessible in any way currently. I would like to do a UI tour but would rather not have to set up website infrastructure to do that!

```
browser.search.geoSpecificDefaults.url
```
I've locked this - the other pref is useful as it means I can set country specific search engines where it's more useful/popular - e.g. Yandex for Russian users (a big demographic!). Hopefully I can move it over to just language packs handling it :-)

PR #1813, opened 2020-11-04
feat: add Startpage search engine choice screen

jamsea:
Adds a screen to prompt the user to set Startpage.com as their default search engine on the first boot up.
Screen Shot 2020-11-02 at 3 10 57 PM


AM4Ryzen:
Why the sus one? Why not DuckDuckGo or Qwant?

hawkeye116477:
@asdf2345 Because Startpage owners are Waterfox's sponsors.

PR #1820, opened 2020-11-12
Follow-up edit of firefox.js v3.

Peacock365:
- Hopefully get rid of all remaining telemetry.
- Hard-disable sponsored content and Pocket on the New Tab Page.
- Remove language pack download URL.
- Change regional search defaults and remove related URL.
- Removed several mobile-related URLs.

Some other findings @MrAlex94, as of Waterfox G3.0.0 Preview, please take a look, some of them are serious:

- Upon startup, Waterfox connects to the following domains: _firefox.settings.services.mozilla.com_ and _content-signature-2.cdn.mozilla.net_ ... According to the following article, the former can't be disabled with a pref: https://support.mozilla.org/en-US/kb/how-stop-firefox-making-automatic-connections - So we probably have to stop it at the code level somehow. Waterfox should not connect to Mozilla aside from DetectPortal, extension updates, updates for the blocklist and updates for the tracking protection.
- Waterfox still advertises Mozilla services under about:protections (scroll down to the bottom of the page)
- The hint at non-existant mobile versions under about:preferences#sync could potentially be removed.
- Under about:license you should state "Waterfox Ltd." instead of just "Waterfox".

Minor ramblings:

- The megabar still exists, and I hate it. There is no reason for the address bar to swell to such a size. Perhaps you think it looks cool, but the criticism it already suffered should tell you otherwise. Please reconsider.
- Opening a new tab for every search is excessive and leads to tab clutter very quickly. This shouldn't be the default, IMHO.

I am very happy with the release so far, but these issues should be looked at. Thank you for your attention.

### DIFF
--- a/browser/app/profile/firefox.js
+++ b/browser/app/profile/firefox.js
@@ -1592,7 +1592,7 @@ pref("toolkit.telemetry.geckoview.maxBatchStalenessMS", -1, locked);
 pref("toolkit.telemetry.geckoview.streaming", false, locked);
 pref("toolkit.telemetry.ipcBatchTimeout", -1, locked);
 pref("toolkit.telemetry.isGeckoViewMode", false, locked)"
-pref("toolkit.telemetry.testing.override.Products.Check", false, locked);
+pref("toolkit.telemetry.testing.overrideProductsCheck", false, locked);
 
 // Enable GMP support in the addon manager.
 pref("media.gmp-provider.enabled", true);
--- a/browser/app/profile/firefox.js
+++ b/browser/app/profile/firefox.js
@@ -39,7 +39,7 @@ pref("extensions.getAddons.cache.enabled", true);
 pref("extensions.getAddons.get.url", "https://services.addons.mozilla.org/api/v3/addons/search/?guid=%IDS%&lang=%LOCALE%");
 pref("extensions.getAddons.search.browseURL", "https://addons.mozilla.org/%LOCALE%/firefox/search?q=%TERMS%&platform=%OS%&appver=%VERSION%");
 pref("extensions.getAddons.link.url", "https://addons.mozilla.org/%LOCALE%/firefox/");
-pref("extensions.getAddons.langpacks.url", "https://services.addons.mozilla.org/api/v3/addons/language-tools/?app=firefox&type=language&appversion=%VERSION%");
+pref("extensions.getAddons.langpacks.url", "", locked);
 pref("extensions.getAddons.discovery.api_url", "", locked);
 
 // Use bloomfilters for the addons blocklist, instead of JSON only.
@@ -390,7 +390,7 @@ pref("browser.helperApps.showOpenOptionForPdfJS", true);
 pref("browser.search.searchEnginesURL",      "https://addons.mozilla.org/%LOCALE%/firefox/search-engines/");
 
 // Market-specific search defaults
-pref("browser.search.geoSpecificDefaults", true);
+pref("browser.search.geoSpecificDefaults", false, locked);
 pref("browser.search.geoSpecificDefaults.url", "", locked);
 
 // search bar results always open in a new tab
@@ -1342,6 +1342,10 @@ pref("browser.newtabpage.activity-stream.discoverystream.spocs.personalized", fa
 
 // User pref to show stories on newtab (feeds.system.topstories has to be set to true as well)
 pref("browser.newtabpage.activity-stream.feeds.section.topstories", false, locked);
+pref("feeds.system.topstories", false, locked);
+
+// No sponsored content on the New Tab Page.
+pref("browser.newtabpage.activity-stream.showSponsored", false, locked);
 
 // The pref controls if search hand-off is enabled for Activity Stream.
 #ifdef NIGHTLY_BUILD
@@ -1572,6 +1576,24 @@ pref("toolkit.telemetry.ecosystemtelemetry.enabled", false, locked);
 pref("browser.ping-centre.telemetry", false, locked);
 pref("browser.ping-centre.log", false, locked);
 
+// No New Tab Page Telemetry.
+pref("browser.newtabpage.activity-stream.feeds.telemetry", false, locked);
+pref("browser.newtabpage.activity-stream.telemetry", false, locked);
+pref("browser.newtabpage.activity-stream.telemetry.structuredIngestion", false, locked);
+pref("browser.newtabpage.activity-stream.telemetry.structuredIngestion.endpoint", "", locked);
+pref("browser.newtabpage.activity-stream.telemetry.ut.events", false, locked);
+
+// Disable various other forms of telemetry.
+pref("privacy.trackingprotection.origin_telemetry.enabled", false, locked);
+pref("telemetry.number_of_site_origin.min_interval", -1, locked);
+pref("telemetry.origin_telemetry_test_mode.enabled", false, locked);
+pref("toolkit.telemetry.geckoview.batchDurationMS", -1, locked);
+pref("toolkit.telemetry.geckoview.maxBatchStalenessMS", -1, locked);
+pref("toolkit.telemetry.geckoview.streaming", false, locked);
+pref("toolkit.telemetry.ipcBatchTimeout", -1, locked);
+pref("toolkit.telemetry.isGeckoViewMode", false, locked)"
+pref("toolkit.telemetry.testing.override.Products.Check", false, locked);
+
 // Enable GMP support in the addon manager.
 pref("media.gmp-provider.enabled", true);
 
@@ -1647,10 +1669,10 @@ pref("browser.contentblocking.report.monitor.home_page_url", "https://monitor.fi
 pref("browser.contentblocking.report.manage_devices.url", "https://accounts.firefox.com/settings/clients");
 pref("browser.contentblocking.report.endpoint_url", "https://monitor.firefox.com/user/breach-stats?includeResolved=true");
 pref("browser.contentblocking.report.proxy_extension.url", "https://fpn.firefox.com/browser?utm_source=firefox-desktop&utm_medium=referral&utm_campaign=about-protections&utm_content=about-protections");
-pref("browser.contentblocking.report.lockwise.mobile-ios.url", "https://apps.apple.com/app/id1314000270");
-pref("browser.contentblocking.report.lockwise.mobile-android.url", "https://play.google.com/store/apps/details?id=mozilla.lockbox&referrer=utm_source%3Dprotection_report%26utm_content%3Dmobile_promotion");
-pref("browser.contentblocking.report.mobile-ios.url", "https://apps.apple.com/app/firefox-private-safe-browser/id989804926");
-pref("browser.contentblocking.report.mobile-android.url", "https://play.google.com/store/apps/details?id=org.mozilla.firefox&referrer=utm_source%3Dprotection_report%26utm_content%3Dmobile_promotion");
+pref("browser.contentblocking.report.lockwise.mobile-ios.url", "", locked);
+pref("browser.contentblocking.report.lockwise.mobile-android.url", "", locked);
+pref("browser.contentblocking.report.mobile-ios.url", "", locked);
+pref("browser.contentblocking.report.mobile-android.url", "", locked);
 
 // Protection Report's SUMO urls
 pref("browser.contentblocking.report.lockwise.how_it_works.url", "https://support.mozilla.org/1/firefox/%VERSION%/%OS%/%LOCALE%/password-manager-report");
@@ -1763,6 +1785,7 @@ pref("extensions.pocket.api", "", locked);
 pref("extensions.pocket.enabled", false, locked);
 pref("extensions.pocket.oAuthConsumerKey", "", locked);
 pref("extensions.pocket.site", "", locked);
+pref("browser.newtabpage.activity-stream.section.highlights.includePocket", false, locked);
 
 // Can be removed once Bug 1618058 is resolved.
 pref("signon.generation.confidenceThreshold", "0.75");
@@ -1799,7 +1822,7 @@ pref("webchannel.allowObject.urlWhitelist", "");
 // Whether or not the browser should scan for unsubmitted
 // crash reports, and then show a notification for submitting
 // those reports.
-  pref("browser.crashReports.unsubmittedCheck.enabled", false, locked);
+pref("browser.crashReports.unsubmittedCheck.enabled", false, locked);
 
 // chancesUntilSuppress is how many times we'll show the unsubmitted
 // crash report notification across different days and shutdown
@@ -1873,7 +1896,7 @@ pref("browser.chrome.errorReporter.infoURL",
 
 // Normandy client preferences
 pref("app.normandy.api_url", "", locked);
-pref("app.normandy.dev_mode", false);
+pref("app.normandy.dev_mode", false, locked);
 pref("app.normandy.enabled", false, locked);
 pref("app.normandy.first_run", false, locked);
 pref("app.normandy.logging.level", 50, locked); // Warn
@@ -2394,3 +2417,6 @@ pref("first-startup.timeout", 30000);
 #ifdef XP_WIN
   pref("default-browser-agent.enabled", false, locked);
 #endif
+
+// Disable extensions recommendations
+pref("extensions.getAddons.showPane", false);


MrAlex94:
Appreciate the combing through, but this isn't the correct way to disable most of those preferences. They have already been disabled [here](https://github.com/MrAlex94/Waterfox/commit/a00c49824401253ce43a579db6f714ad45a2c896#diff-3c7f689d8d070e495e68bcc9f351f5a2d77290ff51644e27598afc170d3863f7).

Also, I'm going to leave `browser.search.geoSpecificDefaults`, it's a useful feature setting the correct search for each geo (even though we don't do it yet). Some countries cost us quite a lot to serve, and others don't benefit at all from having certain search as default. E.g. Yandex is useful in Russia, Bing not so much.

Edit: I will double check about `firefox.settings.services.mozilla.com` and `content-signature-2.cdn.mozilla.net` though.

Peacock365:
@MrAlex94 Ah, I see. OK. 

I am replying though because I've found further information on _content-signature-2.cdn.mozilla.net_, seems like this is beneficial for users:

https://www.reddit.com/r/firefox/comments/hlq14j/automatic_connection_to/fx3638b/
https://github.com/mozilla-services/autograph/blob/master/signer/contentsignature/README.rst

_firefox.settings.services.mozilla.com_ seems to be related to Firefox Experiments, and is probably not desirable for us:

https://wiki.mozilla.org/Mobile/Fennec/Android/Switchboard

So perhaps it would be prudent to limit your effort to blocking _firefox.settings.services.mozilla.com_ somehow...

Peacock365:
@MrAlex94 

Further findings:

- Language packs don't seem to be included yet, I can't find them easily anyway.
- Hamburger menu --> "Sign in to Waterfox" menu still advertises Firefox Monitor & Firefox Send.
- I can't currently install add-ons from AMO, probably due to the custom user agent. We should probably use a hardcoded Firefox 78 ESR user agent for AMO.

Thanks for listening, I hope all the issues I mentioned here so far are being looked at.

Peacock365:
@MrAlex94, are you looking at those problems? I realize that other things like a working transition Waterfox Current --> Waterfox G3 are more important at the moment, but some kind of short reply would be nice, so that I can be convinced that I do not write this stuff in vain. "I am looking at it." would suffice already.

Anyway, I've found an additional conncetion that Waterfox establishes on a regular basis: _aus5.mozilla.org_
Seems to be some kind of Mozilla update server and probably shouldn't be happening anymore. It's not for add-on updates, but for application updates. Please take a look. Thanks in advance.

MrAlex94:
I have seen, but a lot of your commits have a mix of personal preferences and settings that don't need to be changed.

Still, there are useful changes and I will merge when I can.

Also I believe that aus5 connection is for OpenH264 and Widevine download.

Peacock365:
No, I didn't mean my proposed about:config changes at all. I meant the following things:

- _firefox.settings.services.mozilla.com_ should be dealt with.
- Language packs are currently non-functional.
- Hamburger menu --> "Sign in to Waterfox" menu advertises Firefox Monitor & Firefox Send. So does the about:protections page.
- about:license should state "Waterfox Ltd." instead of just "Waterfox" in the first sentences.

Peacock365:
Regarding _firefox.settings.services.mozilla.org_ I have also found this, may yield useful info, also regarding the location of the responsible code:

https://bugzilla.mozilla.org/show_bug.cgi?id=1598562

Peacock365:
@hawkeye116477 may also be interested in these things.

MrAlex94:
I think I've stopped calls to `firefox.settings.services.mozilla.org` with 7170da7685f84.

Peacock365:
That's good to hear. Kind of freaks me out that there is no about:config setting to turn this off, when this should be standard in Firefox. This has me very worried regarding what we will have to deal with once Firefox 91 ESR comes around. We shall see, I guess.

Any progress regarding this cosmetic issue?

- Hamburger menu --> "Sign in to Waterfox" menu advertises Firefox Monitor & Firefox Send. So does the about:protections page.

Peacock365:
I am not sure anymore whether _content-signature-2.cdn.mozilla.net_ is necessary, either. This Tor issue seems to indicate that we could do away with it, as well:

https://gitlab.torproject.org/legacy/trac/-/issues/31575

The comment of Alex Catarineu in this thread seems relevant to me.

**EDIT:** Might also be relevant info: https://github.com/arkenfox/user.js/issues/966

Peacock365:
We should also consider the following:

- Set _network.predictor.enabled_ to _false_.
- Setting the tracking protection to "Strict" by default - what "Strict" does today used to be the "standard" setting a few versions ago, I find it hilarious that the current "standard" tracking protection setting doesn't include tracking content, i.e. trackers.

MrAlex94:
"Firefox Monitor" is useful although may remove Firefox Send. I can't find any reference to `content-signature-2.cdn.mozilla.net`. Are you sure it's still in the source code?

Peacock365:
> "Firefox Monitor" is useful although may remove Firefox Send.

While I agree that Firefox Monitor is useful, it's not built in. The only thing the shortcut does is to link to https://monitor.firefox.com/ ... I think we shouldn't advertise Mozilla services like that, except when they are built in like Sync is. Linking to their products in Waterfox is not that great.

> I can't find any reference to content-signature-2.cdn.mozilla.net. Are you sure it's still in the source code?

about:networking still shows it as a connection after some time. Disclaimer: New Waterfox G3 profile, I didn't update from Waterfox Current.

Peacock365:
Bad news @MrAlex94, it seems like Waterfox G3.0.2 still connects to _firefox.settings.services.mozilla.com_...

Peacock365:
So, in summary, this is what IMHO still needs some fixing:

- _firefox.settings.services.mozilla.com_ and possibly _content-signature-2.cdn.mozilla.net_ connections
- Firefox Monitor & Firefox Send ads in the Sync menus (please...)
- Classic-related, IMHO macOS Big Sur support still needs some of these patches here: https://github.com/MoonchildProductions/UXP/commits?author=dbsoft

hawkeye116477:
In case of Big Sur, this might be also useful => https://www.reddit.com/r/waterfox/comments/k9nnd5/no_sound_for_wf_classic_202012_in_bigsur_while/gf5km9q/?utm_source=reddit&utm_medium=web2x&context=3

Peacock365:
Hi @MrAlex94, some initial findings in Waterfox 3.1.0 (great release so far!):

- Still connects to _firefox.settings.services.mozilla.com_ sadly. In the meantime, I have found out that it is a completetely unneeded connection: https://firefox-source-docs.mozilla.org/browser/components/newtab/content-src/asrouter/docs/ - it seems to me that it can be disabled by editing omni.ja, or so the Arch Wiki says: https://wiki.archlinux.org/index.php/Firefox/Privacy#Editing_the_contents_of_omni.ja
- Please remove references to Firefox Lockwise and Firefox Send. _Please, I implore you._ They are not our services and contrary to Sync, are not an essential part of the browser.
- I tested both the Chrome Web Store and the Opera Store, and was able to download extensions. However, the first click on "Add to Waterfox" didn't seem to do much if anything, the second click after some time immediately added the extension. Weird.
- On macOS, the icons of the stats bar (zoom, fullscreen) get moved to the left of the status bar after a restart for some reason.
- If you plan on doing an Android release, consider using Fennec F-Droid as a base, reason being advanced removal of Mozilla annoyances: https://f-droid.org/en/packages/org.mozilla.fennec_fdroid/
- If you plan on introducing Private Tabs, consider the abandoned "Private Tabs" extension for possible contributions: https://github.com/Infocatcher/Private_Tab
- Consider merging my open commits, I think all changes proposed there are for the benefit of the users. If you have any questions regarding my reasoning, I am happy to answer.

Peacock365:
@dbsoft 

Whatever you did to UXP to make it run on macOS 11.0 Big Sur has clearly worked, I am using your White Star unofficial build of Pale Moon on that OS. However, Waterfox Classic seems to miss your patches. It reproducibly crashes when I hit the green "Maximize" button on ANY window. Is there a possibility of porting those patches you wrote for UXP to Waterfox Classic? Thank you for your attention.

Waterfox Classic 2021.03, macOS 11.2.3 in my case.

dbsoft:
I'll check it out when I get some time, might be a week before I can get to it though... in the middle of some Android code right now.  If you don't hear from me in a week or two send me a message to make sure it didn't slip my mind.

Peacock365:
Thank you! Since you seem to be accustomed to macOS and since you have already written related patches (which might be easily applicable here), I thought I might just approach you with this, because the crashes drive me crazy. :)

Glad that you keep up the macOS builds of Pale Moon despite the way you were treated in their forums, for no visible reason. They should have thanked you for your efforts, but no... Rest assured that what you do is appreciated by people like me and I am sure many others as well! :)

Peacock365:
@dbsoft 

You said that I should remind you of the issue I mentioned re. Waterfox Classic on macOS Big Sur, I am doing that with this post. :-)

Next to the crash that occurs when trying to use WF Classic in fullscreen mode, I found two other issues:

- Waterfox Classic doesn't seem to play any video on macOS 11.0 Big Sur. Try any kind of YouTube video and see for yourself.
- The user agent weirdly says the following: _Mozilla/5.0 (Macintosh; Intel Mac OS X 10.0; rv:56.0) Gecko/20100101 Firefox/56.0 Waterfox/56.5_, however, shouldn't that be "_Mac OS X 10.16_" on Big Sur, going by the Firefox user agent?

Thank you in advance, you seem to have a certain macOS expertise and you are my only hope.

dbsoft:
Ah, yeah I guess Waterfox Classic also needs the versioning changes.... 
https://repo.palemoon.org/MoonchildProductions/UXP/commit/e43b01e58b5477b06736dcaa36979cd0e2fc33f0
and
https://repo.palemoon.org/MoonchildProductions/UXP/commit/b1ac886a988cab94463c265fccb48229d5e1700e

It had only been able to store versions 0 to 15 so 10.16 will loop back to 0. 

Now that I got my White Star fork of Pale Moon out I can take a look at the full screen crash....

Peacock365:
@dbsoft, my hero, I _need_ your help. WF Classic 2021.06 still crashes on macOS Big Sur when I try to run it in fullscreen mode. Or @MrAlex94 could take a look at it...

Peacock365:
@adamp01 this is also for you, please fix this critical bug on macOS Big Sur.

dbsoft:
Sorry I started to look at it, but I ran out of space on my Big Sur Mac.. and wasn't able to do a build... I have a 2012 MacBook Pro which is running Big Sur unofficially... and I have a 2017 MacBook Air and both are out of space... I can probably replace the SSD in the 2012 MacBook Pro pretty easily... just haven't had a chance to do it yet.

Peacock365:
@dbsoft No problem, I merely asked you because you already had comparable patches done for Pale Moon, so that seemed to be the route to a quick fix. I don't blame anyone for not fixing this issue except for the devs @MrAlex94 and @adamp01, who clearly have a machine running macOS Big Sur (I know it because I saw it under about:buildconfig), yet don't seem to care to fix this critical bug of Waterfox Classic. I even linked to someone (you, @dbsoft) who has already provided patches, and still nothing coming from them. Only the main devs are to blame, they are responsible for the project.

dbsoft:
I intend to work on it and see about getting patches into it, just ran into that snag... figured I'd get a replacement SSD and upgrade the system (it is full due to the work I've been doing with iOS and Android and the emulators... and it has been my primary system for that work and after checking out the WaterFox source and starting the build I completely ran out of space) .... but you know how life is... I just haven't gotten around to it... I got the new White Star build done early yesterday (I do White Star builds on an original MacPro running 10.11)... so going to run to the store and see about getting a big SSD for that MacBook Pro today... so if all goes well I might have some news later in the week.

Update: Ok scratch that apparently my model has a proprietary internal connector and I just bought an SSD I can't install into the Mac.

Update 2: I bought an external drive instead and used WinClone to copy my Windows partition to the external drive... allowing me to reclaim 100GB on the internal SSD... just started a build... a bit busy this week but I'll be able to look at stuff at some point.

Update 3: I have successfully built Waterfox Classic and I see what the issue is, trying to figure out the fix now.

Peacock365:
Great to hear! I get the hunch that if you don't fix it, nobody will.

Peacock365:
@dbsoft

Question: Didn't you already have Pale Moon patches ready? Shouldn't those be applicable to Waterfox Classic 1:1?

dbsoft:
@Peacock365  The issue with fullscreen, if it is the same one I am seeing... is not a problem I saw in Pale Moon, and quite frankly it is a bit perplexing. When entering full screen in Waterfox Classic, it hits an undefined selector error... deep inside AppKit... which is quite perplexing, since it isn't actually directly in any Waterfox code. 

```
2021-06-21 08:45:40.562 waterfox[15332:1584234] -[NSView needsOpaqueBackstop]: unrecognized selector sent to instance 0x13e64e400
2021-06-21 08:45:40.563 waterfox[15332:1584234] Mozilla has caught an Obj-C exception [NSInvalidArgumentException: -[NSView needsOpaqueBackstop]: unrecognized selector sent to instance 0x13e64e400]
```

Been trying various things to prevent it from hitting that code path without success, and have been comparing the code to the code in my White Star... but the code is actually almost identical... so I am a bit perplexed at this point... but that is how it is, until you figure it out. ;)

I have patches for the versioning changes... which I'll include too, but trying to figure out the fullscreen issue.

Peacock365:
Thank you so much for all the time and effort you put into this, you are the only hope for Mac users seeing how the main devs don't care. @hawkeye116477 never ever tests for Mac, but rather only Linux, so Mac issues are never tackled by him.

I have identified four bugs on macOS 11.0 Big Sur so far:

- Waterfox Classic crashes whenever someone tries to enter into fullscreen mode (green button at the top left corner).
- The macOS operating system version is not correct in the user agent.
- Waterfox Classic can't play any video IF Waterfox Classic has e10s / multiprocess enabled. Curiously, videos do play when you disable e10s in the settings. You can try this with any YouTube video and a fresh profile, it won't play if e10s is enabled and íf you run macOS 11.0 Big Sur.
- There is a graphical bug if you have a great number of tabs open: There is an empty space at the right side of the window: the tabs + new tab button don't reach the end of the window. Try it yourself with a very large number of tabs (20 or 30 tabs just to make sure) and look at the right side of the tab bar (again, fresh profile, no add-ons).

I found the latter two bugs at a later stage, but the crashing issue when trying to go into fullscreen mode is the most glaring issue. Again, thank you for all your effort, I am just reporting here what I have found using Waterfox Classic on Big Sur.

dbsoft:
@Peacock365 I think I may have narrowed down the problem.  I think it has to do with the use of NSTexturedBackgroundWindowMask and not using  titlebarAppearsTransparent=YES in the ToolBarWindow. 

Waterfox Classic is using the old method of drawing in the entire frame.... White Star (Pale Moon) does the same thing, but we disabled this behavior, so we don't actually use this, which explains why this isn't affecting our browsers.  The problem is that the code to fix this is only available on 10.10 and later, thus it will take some creative conditional coding to get the browser to continue to work on 10.7-10.9 and fix the bug on 11.0.

Peacock365:
Thanks for putting in the effort to identify the issue. Thumbs up for that! This bothers me a lot because I like portable devices (notebooks with 13" screens), but at the same time, due to the limited screen space, I would like to run the browser maximized, but currently can't (crash).

Do you have any idea what causes the "video - e10s bug", and can you confirm the issue I described for the tab bar? Thanks in advance.

Your White Star build has none of these issues, and I like it, but I keep Waterfox Classic around because a mixture of legacy add-ons + WebExtensions seems optimal for me.

Peacock365:
@dbsoft

You seem to have identified the issue, which is great. Any progress on a fix? This issue is really annoying and you are the only hope for Mac users, literally.

dbsoft:
@Peacock365 I do have a fix for it, but it breaks other stuff... and it only works on 10.12 and later.  There may be other workarounds for it... that are less... invasive... because the fix that will work on Big Sur and all the old versions (10.7 to 10.11) will be messy... so kind of trying to find other fixes before investing time on the big messy fix.  The current fix involves switching to the new APIs introduced in 10.12 that the new versions of Firefox are using... but it is a lot of changes and getting it to also work on the old version will have just tons of runtime checks with different code paths.  

Ideally, I would like to just have it avoid the broken code path on Big Sur.. .and keep everything the same on older versions.... but since it seems to be baked into the AppKit code when triggering fullscreen mode without hitting any Waterfox code, I might have to do the big ugly fix.

Issue #1802, opened 2020-10-26
Classic 2020.10 downloads images from Twitter without file extension


AkazaAkarinn:
When using the Save As feature to download images from Twitter, Waterfox does not add the file extension to the image being saved, thus the image is saved as a general file instead of a jpg/png. You can still manually add the file extension to the filename to properly save an image, but this should be done automatically. It doesn't matter what the "Save as type:" is set to either, it only works if you manually add the file extensions. Twitter is so far the only site that I experienced this issue with. 

The best way to save from Twitter is to enlarge a photo and then click and drag that photo to a new tab. 
For example, 
https://twitter.com/tkch870/status/1317746343381671936
becomes
https://pbs.twimg.com/media/EkmTT47UYAANTn1?format=jpg&name=large

I save images using the second link and this is where I've experienced my issues. 

**Expected behavior**
Previously, there was no issue on Classic 2020.09. It is also not an issue on Current 2020.10. I reverted back to Classic 2020.09 and the save as behavior returns to normal. 
On normal behavior, the file extension is included in the filename. However even if you remove the file extension from the filename in the save as dialogue, the file should save with the correct extension listed in the "Save as type:" dialogue. 

**My configuration** 
 - Device: PC
 - OS: Windows 8.1
 - Version: Classic 2020.10

Observerist:
With this latest Classic release 2020.10, image file extensions are scrapped when there is a .php or .cgi behind the file name. With some sites I also get a xxxxxx.jpg.cgi or a xxxxxx.jpg.php. Easier to get rid off then retype the missing extension.

avatar-zero:
Can confirm this bug. 2020.10 appears to be enforcing the filetype extension provided by the URL rather than appending one to custom Save As filenames based on reported MIME type (images from twitter are therefore effectively enforcing a blank extension while script managed servers are effectively enforcing the extension of their respective server-side scripting language).

hawkeye116477:
@MrAlex94 Looks like that reverting  https://github.com/MrAlex94/Waterfox/commit/b48d3d4aff28e4375442bd720b46f72b60988b1e solves that problem.

AkazaAkarinn:
Updating to 2020.12 Fixed the issue.

PR #1828, opened 2020-11-20
[Classic] Backout Bug 1663642 for causing #1802

hawkeye116477:
This reverts commit b48d3d4aff28e4375442bd720b46f72b60988b1e for causing default file extension removed in file picker for save operations.


MrAlex94:
This is a security bug fix though, not sure it's worth the risk?

hawkeye116477:
Alternative way would be to find correct fix for that, if you find it. Seamonkey Team reverted that also.

They said me:
> 
> It was actually in 56 but has been backed out for exactly this. See Bug
> 1344751  comment 26
> 
> Using save page image etc.. did not set a default extension in the file picker
> after this one. I unfortunately don't have access to the actual bug.

Issue #1829, opened 2020-11-20
Windows: Waterfox G3 Npapi plugin not recognized


ThomasCarvello:
we have tried our SIP multimedia softphone with a Javascript API  based on npapi plugin on Waterfox G3 and it didn't work.

our application type  is not recognised and the  plugin is never launched. 

Work fine on Waterfox Classic and Current.

seems to be the same with flash plugin.

neutrino38:
Are NPAPI plugins supported on Waterfox 3.0 ?

neutrino38:
Here are the trace collected on Waterfox 3.0 Windows. I activated the traces as follow:

`SET MOZ_LOG=Plugin:5,PluginNPP:5,PluginNPN:5
`
The traces are:
```
[Parent 16520: Main Thread]: I/Plugin LoadCachedPluginsInfo : Loading Cached plugininfo for NPSPWRAP.DLL
[Parent 16520: Main Thread]: I/Plugin LoadCachedPluginsInfo : Loading Cached plugininfo for nplivevideoplugin.dll
[Parent 16520: Main Thread]: I/Plugin LoadCachedPluginsInfo : Loading Cached plugininfo for npjp2.dll
[Parent 16520: Main Thread]: I/Plugin LoadCachedPluginsInfo : Loading Cached plugininfo for npdeployJava1.dll
[Parent 16520: Main Thread]: I/Plugin LoadCachedPluginsInfo : Loading Cached plugininfo for NPSWF64_18_0_0_209.dll
[Parent 16520: Main Thread]: I/Plugin LoadCachedPluginsInfo : Loading Cached plugininfo for npGatewayNpapi-x64.dll
[Parent 16520: Main Thread]: D/Plugin nsPluginTag::RegisterWithCategoryManager type = application/x-sharepoint, removing = no
[Parent 16520: Main Thread]: D/Plugin nsPluginTag::RegisterWithCategoryManager type = application/x-sharepoint-uc, removing = no
[Parent 16520: Main Thread]: D/Plugin nsPluginTag::RegisterWithCategoryManager type = application/x-econf-plugin, removing = no
[Parent 16520: Main Thread]: D/Plugin nsPluginTag::RegisterWithCategoryManager type = application/x-java-applet, removing = no
[Parent 16520: Main Thread]: D/Plugin nsPluginTag::RegisterWithCategoryManager type = application/x-java-bean, removing = no
[...]

PR #1845, opened 2020-11-26
Ives/fix remove plugins limitation

neutrino38:
Fix NPAPI limtation on G3.

Fix issue https://github.com/MrAlex94/Waterfox/issues/1829


MrAlex94:
Thank you! Will be in the next release of G3, G3.0.2.

Issue #1843, opened 2020-11-26
G3 fails to launch on El Capitan & ruins user profile for older versions


BezzleBedeviled:
Normally this wouldn't be a problem (nobody expects new software to always run on legacy OSes), except that Waterfox Current's immediately prior versions not only don't ask (by default) before upgrading, but fail to test/sandbox the launch.

GJRobert:
+1
Waterfox G3 officially requires only macOS 10.10, but it does not run on my El Capitan either.

Message when running from command line:
```
Minimum OS version requirement not met!
```

My profile from Classic remains intact, though.

GJRobert:
Duplicate of #1839

GJRobert:
Duplicate of #1839

PR #1871, opened 2020-12-17
[G3] Compatibility fixes

### LINKED ISSUE #1839: Browser quits no error or warning: Mac OS X 10.11.6
Mac OS X 10.11.6
Automatic upgrade broke Waterfox - opens then quits - no crash report, I have lost all access to stored bookmarks etc -
Note: Installing new download of Waterfox G3.0.0.1 Setup (Waterfox 78.5) just does not work, it does the exact same thing...Tried on 2 different MacOS X systems running 10.11.6.

Browsers are supposed to be stable... this is not.  I will have to go back to old copy, Waterfox Current (Version 68.0).

aeiouaeiouaeiouaeiouaeiouaeiou:
Fixes issues #1839 and #1843.

### DIFF
--- a/browser/app/macbuild/Contents/Info.plist.in
+++ b/browser/app/macbuild/Contents/Info.plist.in
@@ -237,7 +237,7 @@
 	<key>LSFileQuarantineEnabled</key>
 	<true/>
 	<key>LSMinimumSystemVersion</key>
-	<string>10.9.0</string>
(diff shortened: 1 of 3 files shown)


GJRobert:
Hi, when will this fix affect the official release build?
AFAIK, v78.7.0 of G3 still does not launch on macOS 10.11.

Issue #1868, opened 2020-12-16
[G3] JS array windows.navigator.plugins does not list any plugin


neutrino38:
**Describe the bug**

The javascript object window.navigator.plugins used to list browser plugins. As documented in 

https://developer.mozilla.org/en-US/docs/Web/API/NavigatorPlugins/plugins

With Waterfox G3, window.navigator.plugins returns an empty array.

**To Reproduce**
 PluginArray { length: 0 }

**Expected behavior**

PluginArray should list all installed and active plugin with their description and version.

**Screenshots**
![image](https://user-images.githubusercontent.com/11951413/102365590-01b36b00-3fb8-11eb-842e-badf5207ba7a.png)

**My configuration** 
 - Device: Windows 10 PC
 - OS: Windows 10
 - Version: Waterfox G3

**Additional context**
I checked the code in dom/base/nsPluginArray.cpp and this seems to be implemented. So for any reason, the nsPluginArray is not listing the plugins correctly.

neutrino38:
I created a partial fix. It is restoring window.navigator.plugins at init. However I am not sure that the nsPluginArray gets updated if a new plugin is detected while Watefox is running or if the page is not refreshed. But this is a rare cornercase that can be ignored by now.

https://github.com/MrAlex94/Waterfox/pull/1875

neutrino38:
Closed : fix ,merged by Alex

PR #1875, opened 2020-12-19
Fix window.navigator.plugins

neutrino38:
Fixing issue https://github.com/MrAlex94/Waterfox/issues/1868


MrAlex94:
Many thanks again for the work!

PR #1903, opened 2021-01-10
[G3] Add few improvements to statusbar

hawkeye116477:
Just like on commits messages.

Zoom field editable means that field which shows current zoom percent can transform to input field when clicking on it and you can type percent value just like on Vivaldi browser.

### DIFF
--- a/browser/app/profile/firefox.js
+++ b/browser/app/profile/firefox.js
@@ -2417,7 +2417,7 @@ pref("browser.tabs.copyallurls", true);
 // 0 - hide Status Bar
 // 1 - show Status Info Overlay
 // 2 - show Status Bar
-pref("browser.statusbar.mode", 0);
+pref("browser.statusbar.mode", 1);
 
 // Don't use buttons in zoom range control in Status Bar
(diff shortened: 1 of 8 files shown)


browser/app/profile/firefox.js
```diff
@@ -2417,7 +2417,7 @@ pref("browser.tabs.copyallurls", true);
 // 0 - hide Status Bar
 // 1 - show Status Info Overlay
 // 2 - show Status Bar
-pref("browser.statusbar.mode", 0);
+pref("browser.statusbar.mode", 1);
```
MrAlex94, line 2420: Do you have an image of what this will look like when in use?

MrAlex94 (approved):

browser/app/profile/firefox.js
hawkeye116477: Not made.
Anyway, overlay is just like pop-up and it appears only when needed, status bar is displayed always and 'hide' option causes that no status is displayed (no overlay and no status bar).

PR #1934, opened 2021-02-05
[G3] Add fixes for Status Bar and company

hawkeye116477:
Colors should be good now for all themes and zoom controls should be on right side by default.


MrAlex94:
Awesome, thank you!

Issue #1980, opened 2021-03-07
[Classic] prefer Light/Dark Theme setting


c33s:
any chance to add `prefers-color-scheme` to waterfox classic?

https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme

jabcreations:
I just commented on this thread:
https://github.com/MrAlex94/Waterfox/issues/1996#issuecomment-803464044

hawkeye116477:
Yes, I'm working on it. Linux part done, testing now, will check W10 soon too.

![2021-03-25_11-53](https://user-images.githubusercontent.com/19818572/112461906-59e8ce80-8d58-11eb-8cd4-75da780c7caa.png)

jabcreations:
Absolutely wild! Please feel free to test dark mode on my site for further validation: https://www.jabcreations.com/
Thank you for your efforts! 😍︀😍︀😍︀

reallyuniquename:
@hawkeye116477
Would `ui.systemUsesDarkTheme` config entry be supported as well?

hawkeye116477:
@reallyuniquename Yes.

jabcreations:
I just updated to Waterfox 56 2021.03 and dark mode isn't working. Additionally I tested window.matchMedia('(prefers-color-scheme: dark)').matches and it returns false. What is going on?

hawkeye116477:
@jabcreations Preference `privacy.resistFingerprinting` must be set to `false` to make it working.

jabcreations:
@hawkeye116477 It _is_ set to `false`. Anything else?

hawkeye116477:
@jabcreations What's your OS?

jabcreations:
Windows 10 Pro x64.

hawkeye116477:
So maybe you got wrong build. I saw that few days ago was hidden 2021.03 installer and that was build before adding prefers-color-scheme feature. Go to `about:buildconfig` and check if you have `Built from https://github.com/MrAlex94/Waterfox/commit/5d6e36f2f62aee766e62e19be21793334ed7c9c7`. If not, then try to download installer again. Otherwise maybe try in safe mode/new profile.

jabcreations:
I also tested a new profile earlier and forgot to mention that it still didn't work. I just restarted with addons disabled and it still doesn't work. Here is the full page dump of about:buildconfig, hopefully this helps:

```
about:buildconfig
Source

[...]

Issue #1937, opened 2021-02-06
WF Classic unable to display site


Gebarx:
WF Classic 2021.01.1 unable to display site [https://www.translatum.gr/forum/](https://www.translatum.gr/forum/) shows blank page. Same with add-ons disabled. Root site [https://www.translatum.gr/forum/](https://www.translatum.gr/) displays OK. The forum displays fine in all other browsers (Firefox, Opera, Chrome, Edge), even in WF Current.

Squall-Leonhart:
requires an unsupported polyfill that is never coming to classic.

roman6626:
There is no secret or miracle here. Waterfox Classic works worse and worse in the modern web. It is technically outdated. And soon this version will just die. There will be only Waterfox Current - practically, the same Firefox.

hawkeye116477:
Appears that problem is with this:
```
if (window.matchMedia("(prefers-color-scheme: dark)").media === "not all") {
    document.documentElement.style.display = "none";
    document.head.insertAdjacentHTML(
        "beforeend",
        '<link rel="stylesheet" href="https://www.translatum.gr/forum/Themes/CustomSunRise/css/sunrise.css?fin20" onload="document.documentElement.style.display =">'
        );
    }
```
In short problem is that Waterfox Classic doesn't support `prefers-color-scheme`, so content is hidden, because webmaster made a bug in code (onload in css is wrong).

If you have uBO or AdGuard, then you can add to them => `translatum.gr##+js(remove-attr, style, html[style="display: none;"])` to fix that.

Alternative solution:
Create userscript in Violentmonkey/Greasemonkey/Tampermonkey
```
// ==UserScript==
// @name        Fix translatum.gr
// @namespace   Violentmonkey Scripts
// @match       https://www.translatum.gr/forum/
// @grant       none
// @version     1.0
// @author      -
// @description 2/8/2021, 1:08:41 PM
// @run-at document-idle
// ==/UserScript==

document.documentElement.style.display = "";
```

PR #2016, opened 2021-03-26
[Classic] Add support for CSS prefers-color-scheme media feature

hawkeye116477:
Tested and works great on W10 and Linux.

In case of Mac OS, these needs to be implemented:
https://bugzilla.mozilla.org/show_bug.cgi?id=1466335
https://bugzilla.mozilla.org/show_bug.cgi?id=1471950
https://bugzilla.mozilla.org/show_bug.cgi?id=1528960
to make it working.

There is also small problem on W10, application window is always light on Default theme, so it's inconsistent with websites.

It solves #1980.

It's workaround for #1937 (I think that's still should be reported to webmaster, who made bug in code).


MrAlex94 (approved):

aeiouaeiouaeiouaeiouaeiouaeiou:
Porting 1466335 is not easy because of significant difference in UI code.

PR #2019, opened 2021-03-30
[Classic] Let PerformanceObserver ride the trains.

hawkeye116477:
It's not all, some tests will still fail, but seems that should suffice for GitHub now :smile:


MrAlex94 (approved):
Great work, thanks.

PR #2021, opened 2021-04-01
update Dockerfile for classic Linux ci

adamp01:


hawkeye116477:
Seems that some things can be little simplified.
You can get llvm-6.0 and clang-6.0 from xenial-updates repo. There is also nasm-mozilla package in xenial-updates universe repo.

Issue #1774, opened 2020-09-28
Cannot browse any kb.vmware.com article, Waterfox show a black page.


sebma:
**Describe the bug**

Hi,
Cannot browse any kb.vmware.com article, Waterfox show a black page.
Example : https://kb.vmware.com/s/article/2005129

**To Reproduce**

1. Go to https://duckduckgo.com/?q=site%3Ahttps%3A%2F%2Fkb.vmware.com&ia=web
2. Click on any result

**Expected behavior**

Waterfox shows the content of the page

**Screenshots**

![kb vmware com_Waterfox_Issue_#1774](https://user-images.githubusercontent.com/4569306/94448353-15b2b200-01ab-11eb-8371-3b5c657e20b9.png)

**My configuration** 
 - Device:  PC
 - OS:  Windows 7 64bits
 - Version: Classic 2020.09 64bits

**Additional context**

hawkeye116477:
Mozregression results:
Last bad: 2018-07-04
First good: 2018-07-05
Pushlog: https://hg.mozilla.org/mozilla-central/pushloghtml?fromchange=6c0fa9a675c91390ca27664ffb626c56e8afea4d&tochange=afdeb0288690f0acde2ba6bd008f860e1acdd026

I guess that's probably same problem as #1252 and few others.

PR #2051, opened 2021-04-19
[Classic] Add few useful user-agent overrides.

hawkeye116477:
Some websites are blocking access, cuz they are checking browser version and because WF Clasic is based on FF 56, they are assuming that it's insecure. So we need to tell them that our browser is super-secure, by telling them that's Firefox 99 :rofl: 

https://www.reddit.com/r/waterfox/comments/mkgovt/error_1020_in_with_waterfox/

I also added override for VMWare, cuz apparently they made site looking great in IE 11, but for Firefox they require `window.event` support. However site sometimes still has problems with viewing (css), but at least articles are readable.

It's workaround for #1774.


MrAlex94:
Hah thank you, much appreciated.

Issue #2061, opened 2021-04-23
Port Blocking Issue


B-IT:
WTF !!!  
Shitty Port blocking !!!
Back to old Version !!

Please Remove the Port Blocking from Version 2021.04.1  !!

wkuzma:
How to ublock port 3000

wkuzma:
I tried to add network.security.ports.banned.override as a string with port 3000  but it does not work.

Peke:
Reverted back to 2021.03 and blocked updates till this is fixed. using network.security.ports.banned.override ; 1-65535 did not work.

Peke:
Duplicate of #2057

wkuzma:
where can I get a copy of  2021.03 to revert back to?

Owdjim:
Yep - completely blocked the world-wide network of KiwiSDR ham radios which use port 8073. network.security.ports.banned.override appears to be ignored. Also blocked from managing any device on my own LAN for heavens sake!

wkuzma:
I found and reverted to Waterfox 2021.02 Release no my necessary port 3000 is no longer blocked for myself 
of Course I have no idea if others are blocked from my web site at port 3000 once they do the update

goofyas007:
Second that, override is ignored.
Please fix ASAP, like mentioned above, local LAN services are not accessible.

justsonic:
I had the same problem - the port 83 is restricted:

> This address is restricted
> This address uses a network port which is normally used for purposes other than Web browsing. Waterfox has canceled the request for your protection.

Adding ports to network.security.ports.banned.override parameter does not help.
I install the old version of Waterfox Classic and disable updates, but it still updating and stops working.

HELP!!!!

W-Geronius:
same problem on port 8080 within my local network (!), which definitely is a standard HTTP port that must not be blocked.
Waterfox classic 2021.04.1 (64-Bit) - found no solution other than installing 2021.03
Please fix !

Killogy:
I don´t know if related to this port problem but I can´t play Acestream stream on ArenaVision https://arenavision.site . Reinstalling Waterfox 2021.03 fixed the problem. 

Stream load from http://127.0.0.1:6878

waltrothfuss:
[...]

PR #2062, opened 2021-04-23
[Classic] Emergency fix

### LINKED ISSUE #2057: Waterfox 2021.04.01 port Bug: This address is restricted.
When trying to access localhost:3000 I get

"This address is restricted. This address uses a network port which is normally used for purposes other than Web browsing. Waterfox has canceled the request for your protection."

Searching the web I got the solution to add "network.security.ports.banned.override" value "3000" to about:config. But it didn't worked.

aeiouaeiouaeiouaeiouaeiouaeiou:
#2057 and #2061.


hawkeye116477:
https://repo.palemoon.org/MoonchildProductions/UXP/commit/5e705bd5059da5b7a39e3a096b7c7f59cb466730

Tampa:
This should really make it into a hotfix, because as it stands these port blocks are breaking content delivery of various platforms and make it impossible to use Waterfox in local networks where things might live on different ports to avoid conflicts.

If this is a wanted feature port from Firefox ideally it should obey some form of debug setting to disable it or come with a similar dialog to risk-acceptance of out of date certs.

PR #2085, opened 2021-05-13
[classic] remove -otmp/ from second 7z call as duplicating tmp dir

adamp01:


MrAlex94 (approved):
My name is Alex and I approve of this commit.

PR #2110, opened 2021-06-02
[Classic] Widevine and certificates update

aeiouaeiouaeiouaeiouaeiouaeiou:
- Update timezone and CA data using [scripts from TenFourFox](https://github.com/classilla/tenfourfox/blob/master/104fx_upcerts.sh). NSS library also should be updated (currently one does not support CKA_NSS_SERVER_DISTRUST_AFTER), but I haven't noticed any serious problems.
- Google discontinued support for all older versions of Widevine: https://bugzilla.mozilla.org/show_bug.cgi?id=1706393


MrAlex94:
Thank you 👍

PR #2131, opened 2021-06-14
[Classic] Add fixes for -moz-appearance and -webkit-appearance

### LINKED ISSUE #2128: Waterfox classic crashes on https://ec.cando-web.co.jp/
```

### Describe the bug
Waterfox classic crashes on https://ec.cando-web.co.jp/ .
Browser also crash with New profile, Safe-mode as well as  Disabling HWA.
in case of e10s is enabled: tab crashes immediately.
in case of e10s is disabled: when mouse over <select> tag  `すべてのカテゴリ   ▼`

### To Reproduce
1. Go to https://ec.cando-web.co.jp/
2. Mouse hover over <select> tag `すべてのカテゴリ   ▼`

### Expected behavior
Nothing crash

### System (please complete the following information):
 - OS: Windows,10 Home 21H2
 - Version 2021.06
```

hawkeye116477:
This fixes #2128 and other problems with -moz-appearance and -webkit-appearance.

[Bug 1500423](https://bugzilla.mozilla.org/show_bug.cgi?id=1500423) is partly done, Mac OS part is missing, I'm leaving that for Mac OS lovers, but in case of Outlook site that problem doesn't appear now.

I tested all on Linux and W10 and seems that all works as intended.

I don't have Mac OS, so I'm leaving testing that part for you.


MrAlex94:
Tested on macOS, works 👍 great job.

Issue #2045, opened 2021-04-16
Google voice breaks on 2021.03 (64-bit) Classic


code-issues-pull:
voice.google.com

Applied the 2021.03 (64-bit) update this morning and since then google voice fails to render properly. I've tried on my main profile and an alternate profile that has zero addons.

Changing user agent does not resolve the issue either.

Console output has a LOT of this happening:  
> Error: The expression cannot be converted to return the specified type.
eka/<@https://www.gstatic.com/_/voice/_/js/k=voice.voice_module_set.en_US.R9smC-bCmL4.O/am=RA/d=1/ct=zgms/rs=AJAOlGme1pci35hid7HzcsC1v639FndlcA/m=base:363:169
Vb/<@https://www.gstatic.com/_/voice/_/js/k=voice.voice_module_set.en_US.R9smC-bCmL4.O/am=RA/d=1/ct=zgms/rs=AJAOlGme1pci35hid7HzcsC1v639FndlcA/m=base:44:216
_.f.eV@https://www.gstatic.com/_/voice/_/js/k=voice.voice_module_set.en_US.R9smC-bCmL4.O/am=RA/d=1/ct=zgms/rs=AJAOlGme1pci35hid7HzcsC1v639FndlcA/m=base:2090:31
_.f.subscribe@https://www.gstatic.com/_/voice/_/js/k=voice.voice_module_set.en_US.R9smC-bCmL4.O/am=RA/d=1/ct=zgms/rs=AJAOlGme1pci35hid7HzcsC1v639FndlcA/m=base:2089:358
Sba.prototype.call@https://www.gstatic.com/_/voice/_/js/k=voice.voice_module_set.en_US.R9smC-bCmL4.O/am=RA/d=1/ct=zgms/rs=AJAOlGme1pci35hid7HzcsC1v639FndlcA/m=base:2194:72
_.f.subscribe@https://www.gstatic.com/_/voice/_/js/k=voice.voice_module_set.en_US.R9smC-bCmL4.O/am=RA/d=1/ct=zgm[…] <gv-side-nav expanded="true" embedded="true">

code-issues-pull:
Still broken on 2021.04.1 (64-bit)

code-issues-pull:
Still broken on 2021.04.2 (64-bit)

jiangzhenjerry:
[...]

PR #2166, opened 2021-07-17
[Classic] Do not throw from PerformanceObserver.observe when none of the entryTypes are known (log a JS console warning instead)

hawkeye116477:
Only Polish language translation is done, others are in English (it would be good if language packs will be finally on some localization management platform like Crowdin or Mozilla Pontoon), but it's message only on console, so it's not that important.

From someone's log from Reddit looks like that may be related with #2045. Anyway, it's geolocked, so I can't check that.

Tests:
http://wpt.live/performance-timeline/po-observe.any.worker.html
http://wpt.live/performance-timeline/po-observe.any.html

-------------------------
Update: I got [confirmation](https://github.com/MrAlex94/Waterfox/issues/2045#issuecomment-882041563).


hawkeye116477:
@MrAlex94 What about this? It's regression from previous versions, so it's the most important :smile:.

MrAlex94 (approved):

PR #2181, opened 2021-07-24
[Classic] Initial config for GitHub Actions builds on classic branch

aeiouaeiouaeiouaeiouaeiouaeiou:
GitHub Actions is completely free and unlimited for open source projects. I think it would be a good idea to add support for this on classic branch to simplify the testing of pull requests (as similar projects do).

- Unfortunately there is no Windows virtual environment on the list, because it is not easy to prepare dependencies there.
- Trigger on "version_display.txt" is needed to checkup branch and rebuild virtual machine cache.
- Artifacts can be disabled via ENABLE_ARTIFACTS_MODE if required.

### DIFF
--- a/.github/workflows/classic.yml
+++ b/.github/workflows/classic.yml
@@ -1,0 +1,103 @@
+name: Build Classic
+
+on:
+    push:
+        paths:
+          - 'browser/config/version_display.txt'
+    pull_request:
+
+jobs:
+    build:
+        strategy:
+            fail-fast: false
+            matrix:
+                os: [macos-10.15, ubuntu-20.04]
+
+        env:
+            ENABLE_ARTIFACTS_MODE: "true"
+            MOZCONFIG: "./.mozconfig-ga-classic"
+            SHELL: "/bin/bash"
+
+        runs-on: ${{ matrix.os }}
+        steps:
+          - name: Set up Xcode version
+            uses: maxim-lobanov/setup-xcode@v1
+            with:
+              xcode-version: '11.2.1'
+            if: startsWith(matrix.os, 'macos')
+
+          - name: Set up Python version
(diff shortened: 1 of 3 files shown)


MrAlex94:
I've looked at GH Actions before, but never went with them due to the limited minutes (would usually only be good for about 10 builds a month, so 3 for each platform). 

I've never found any reference on the actual GH website regarding unlimited for FOSS software, but I have seen websites mention it. The only usage limits I've found are [here](https://docs.github.com/en/actions/reference/usage-limits-billing-and-administration), and they don't seem to have any more information. Do you know where else they might mention it?

hawkeye116477:
It's also on https://docs.github.com/en/billing/managing-billing-for-github-actions/about-billing-for-github-actions.

However, it's mentioned that:

> GitHub Actions usage is free for both public repositories and self-hosted runners. For **private** repositories, each GitHub account receives a certain amount of free minutes and storage, depending on the product used with the account.

And on https://github.com/features/actions you have written that minutes per month limit only applies to private repositories.

MrAlex94:
🤦‍♂️ Looks like you're right. On top of that, looks like unlimited build minutes as well. I will run a few test builds and see how long they take. Free tiers on Travis CI (when it was free for OSS) and AppVeyor would time out at ~90mins. Even if we have unlimited minutes, don't really want to have builds that take over an hour and a half.

hawkeye116477:
@aeiouaeiouaeiouaeiouaeiouaeiou
> Unfortunately there is no Windows virtual environment on the list, because it is not easy to prepare dependencies there.

Seems it's not that hard as you said. I learned that Windows also has great package managers and some of needed things are already installed too. I'll send PR later for that :smile:.

Issue #2032, opened 2021-04-08
[Classic] Bad MP4 file crashes tab


Vulpovile:
this video causes a crash in waterfox classic: https://tornadus.net/assets/video/how.mp4
Steps to reproduce:
Open waterfox classic
Open link
Crash

Squall-Leonhart:
Format                      : MPEG-4
Format profile              : Base Media
Codec ID                    : isom (isom/iso2/avc1/mp41)
File size                   : 457 KiB
Duration                    : 3 s 709 ms
Overall bit rate mode       : Variable
Overall bit rate            : 1 010 kb/s

Video
ID                          : 1
Format                      : AVC
Format/Info                 : Advanced Video Codec
Format profile              : High@L6.2
Format settings             : CABAC / 4 Ref Frames
Format settings, CABAC      : Yes
Format settings, Reference  : 4 frames
Codec ID                    : avc1
Codec ID/Info               : Advanced Video Coding
Duration                    : 3 s 405 ms
Source duration             : 3 s 438 ms
Bit rate                    : 946 kb/s
Width                       : 42 pixels
Original width              : 12 000 pixels
Height                      : 42 pixels
Original height             : 12 000 pixels
Display aspect ratio        : 1.000
Frame rate mode             : Variable
Frame rate                  : 41.307 FPS
Minimum frame rate          : 8.263 FPS
Maximum frame rate          : 60.000 FPS
Color space                 : YUV
Chroma subsampling          : 4:2:0
Bit depth                   : 8 bits
Scan type                   : Progressive
Bits/(Pixel*Frame)          : 12.985
Stream size                 : 389 KiB (85%)
Source stream size          : 397 KiB (87%)
Writing library             : x264 core 161 r3027 4121277
[...]

Issue #1863, opened 2020-12-10
Waterfox classic crashes when open AAC Bitstream


alice0775:
**Describe the bug**
Waterfox classic, tab crashes when open  AAC Bitstream

Build ID: 20201208163116
User Agent : Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:56.0) Gecko/20100101 Firefox/56.0 Waterfox/56.3

**To Reproduce**
1. Open https://www2.iis.fraunhofer.de/AAC/multichannel.html
    or direct open https://www2.iis.fraunhofer.de/AAC/7.1auditionOutLeader_v2_rtb.mp4

**Expected behavior**
No crash

**Screenshots**
NaN

**My configuration** 
 - Device: PC
 - OS: Windows 10 Home version 1909 64bit
 - Version: Classic 2020.12

**Additional context**
Regression window:
  Good: Waterfox56.2.13
  Bad: Waterfox Classic 2020.01.1

hawkeye116477:
It's probably this => https://bugzilla.mozilla.org/show_bug.cgi?id=1509875, 2/3 parts already done, only updating libcubeb left.

PR #2182, opened 2021-07-24
[Classic] Fix media issues

### LINKED ISSUE #2118: Drop in frames on Twitch player every 2-3 seconds
**Describe the bug**
Unless Waterfox is running in Safe mode, it appears that there a few frames dropped on twitch.tv when watching streams. It occurs consistently and repeatedly. I've started a brand new profile with default addons, it didn't fix the issue. I disabled all addons, it did not fix issue. Waterfox version 2021.04.2 although this has happened in previous versions as well. 
*IF THIS BUG IS LIKELY REPRODUCIBLE IN FIREFOX, PLEASE REPORT TO MOZILLA.*
Not reproducible in Firefox nor IceCat, it appears to be confined to Waterfox

**To Reproduce**
1. Go to 'twitch.tv'
2. Click on 'any stream'
3. Scroll down to 'n/a'
4. See error when watching streamers

**Expected behavior**
No drop in frames and having a smooth viewing experience

**Screenshots**
I could add a video capture of the Waterfox window if that will help.

**System (please complete the following information):**
 - OS: Win10
- Waterfox version 2021.04.2 - Waterfox Classic

### LINKED ISSUE #2179: Waterfox classic 2021-7 crashes in this site
### What happened?

waterfox crashes on https://www.solasta-game.com/ - not only the tab where the site is loaded crashes but all other opened tabs crash too. The site works flawlessly on firefox.

### Reproducible?

- [X] I have tested this on Firefox.

### Version

G4.𝑋 (Future)

Windows

### Relevant log output

_No response_

hawkeye116477:
This fixes #2179, #2032 and #1863 for Windows.

After that it should show error message instead of crashing.

In case of #1863, audio file don't play, there is no visual error message, it's only on console. To make it playable updating libcubeb is needed, but it can be not that straightforward.

This also fixes #2118 for all OSes.


Felipefpl:
Great! 5 fixes into 1, i hope it'll be submitted. :)

MrAlex94 (approved):

PR #2183, opened 2021-07-26
[WFX-173] Introduce new background script to extensibles to allow ren…

adamp01:
…dering additional preference checkboxes on about:preferences

### DIFF
--- a/browser/extensions/extensibles/_locales/en/messages.json
+++ b/browser/extensions/extensibles/_locales/en/messages.json
@@ -24,10 +24,10 @@
     "message": "Status Bar"
   },
   "copyUrl": {
-    "message": "Copy URL"
+    "message": "Copy Tab Link"
   },
   "copyAllUrls": {
-    "message": "Copy All URLs"
+    "message": "Copy All Tab Links"
   },
   "restartBrowser": {
     "message": "Restart"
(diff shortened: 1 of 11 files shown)


MrAlex94:
All looks good. Just a couple of UI/X changes to make it more inline with current vernacular throughout the UI:

- Disable `Copy URL only from active tab` by default
- Use `Link`/`Links` instead of `URL`/`URLs`

URL is definitely the correct usage, but since the browser and the average user uses link in it's place, might as well keep it consistent with that.

MrAlex94 (changes requested):
All looks good. Just a couple of UI/X changes to make it more inline with current vernacular throughout the UI:

- Disable `Copy URL only from active tab` by default
- Use `Link`/`Links` instead of `URL`/`URLs`

URL is definitely the correct usage, but since the browser UI and the average user uses link in it's place, might as well keep it consistent with that. I know hawkeye implemented it with URLs but might as well fix it up now before we forget 😬

hawkeye116477:
`I know hawkeye implemented it with URLs`

In principle, I borrowed this from Cyberfox and didn't changed author's entries. Anyway, you're right.

MrAlex94 (approved):

PR #2185, opened 2021-07-26
[Classic] Update user-agent overrides

hawkeye116477:
More and more websites are closing the doors for Classic with message like „Please update your browser", so I think that the best would be to do same like Moonchild did, so set 68.0 user-agent.

To make it more comfortable for the future to change compatibility version I borrowed code from Moonchild.
Now UA compatibility version can be read from pref (`general.useragent.compatMode.version`). There is also possibility to toggle short/long Waterfox version string in user-agent (`Waterfox 56.5.2021.07` or `Waterfox 2021.07`).

Default user-agent string looks like this: `Mozilla/5.0 (X11; Linux x86_64; rv:68.0) Gecko/20100101 Firefox/68.0 Waterfox/56.5.2021.07`.

I also updated site specific user-agents (some have been removed, few changed, few added, borrowed from Pale Moon and Basilisk).


Felipefpl:
1st of all thank you for this change but there is something that came to my mind:

Firefox now is at version 90 and somewhat close to 91, how long will this last till version 68 is considered outdated by those same sites?

hawkeye116477:
Well, 68 is well-tested by Moonchild, it's definitely better than 56.

In the future maybe when more things will be ported, then maybe another bump would be good.

Felipefpl:
Got it, thank you for the explanation, i hope 68 will last at least for a little bit.

MrAlex94 (approved):

hawkeye116477:
It's approved, but not merged. What's the reason?

MrAlex94:
> It's approved, but not merged. What's the reason?

I was traveling at the time, and lost Internet so could not approve until now :-)

PR #2210, opened 2021-08-13
[Classic] Compile with a C++14-compatible mode and cleanup a little

hawkeye116477:
I'll need C++17 for new regexp features. For now let's try to solve C++14 issues.

We'll see how it's on Mac ~and I'll check later on W10~.


hawkeye116477:
It's ready to merge now. All currently supported distros has compilers with C++14 support, so switching to it shouldn't be problem.

MrAlex94 (approved):

PR #2238, opened 2021-09-06
[Classic] Add Windows builds for testing

hawkeye116477:
Thanks to scoop, it's easy to install depends, but unfortunately building takes about 2 hours.

PS: If you will want someday newer version of MozillaBuild, then `mozilla-build.json` file will need to be edited of course .

### DIFF
--- a/.github/workflows/classic.yml
+++ b/.github/workflows/classic.yml
@@ -4,18 +4,19 @@ on:
   pull_request:
     branches:
       - classic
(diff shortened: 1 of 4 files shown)


hawkeye116477:
>  if caching has been removed from .mozconfig-ga

It's not removed.

MrAlex94 (approved):

MrAlex94 (approved):

PR #2241, opened 2021-09-08
[Classic] Add Docker image for building on GH Actions

hawkeye116477:
Thanks to that you could build on CentOS 7 and lower requirements of GNU libc (glibc/libc6) to 2.17.

It's set to build Docker image at 00:00 on 20 day-of-month and it's possible to trigger manually through Actions tab.

Maybe if you will add installing few additional packages, then that could be used also for Waterfox G3.

PS: I set 7.0 as LLVM toolset version, but if you want newer, then you need to uncomment commands which adds Springdale Linux repo (which also works great on CentOS), which contains newer LLVM toolset versions.
Setting LD_LIBRARY_PATH is needed, because otherwise Clang complains that can't find LLVM libs.

If you will trigger building that after merging, then I will send next PR with changed config for building Classic.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -1,0 +1,48 @@
+name: Build Docker image
+
+on:
+  workflow_dispatch: null
+  schedule:
+    - cron: "00 00 20 * *"
+
+jobs:
+  docker:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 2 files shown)


hawkeye116477:
@MrAlex94 What about this? It should greatly improve compatibility and even Debian 9 users will be able to use official package :smile:.

PR #2242, opened 2021-09-09
[WFX-220] Add release job to production.yml

adamp01:


.github/workflows/production.yml
```diff
@@ -129,9 +130,21 @@ jobs:
...
+              ./objdir-*/dist/update/update.xml
+
+        - name: Upload to CDN
+          run: |
+            rsync -e ssh ./objdir-*/dist/install/sea/*.exe ${{ secrets.FTP_USER }}@${{ secrets.FTP_HOST }}:/92790/waterfox/releases/win64/installer
+            rsync -e ssh ./objdir-*/dist/update/*.mar ${{ secrets.FTP_USER }}@${{ secrets.FTP_HOST }}:/92790/waterfox/releases/win64/update
+
+        - name: Upload new update.xml to test
+          run: |
+            aws s3 cp --content-type="text/xml" --metadata-directive="REPLACE" s3://aus.waterfox.net/update/test/WINNT/update.xml \
```
adamp01, line 145: We are doing this step so that we have a reference to the old version number - we need this so that in the release stage later on we can create a directory for the old version.

PR #2283, opened 2021-10-02
[Classic] Improve performance and update regexp engine

hawkeye116477:
For example:
Bug 1398751
Before:
  stringFromCharCode: 5514 ms.
   stringFromCodePoint: 12928 ms.
After:
 stringFromCharCode: 932 ms.
 stringFromCodePoint: 1127 ms.

RegExp results from Octane benchmark:
Before: 1677
After: 4523

This PR also removes old regexp engine.


MrAlex94 (approved):

ipc/testshell/XPCShellEnvironment.cpp
```diff
@@ -55,7 +55,7 @@ namespace {
 static const char kDefaultRuntimeScriptFilename[] = "xpcshell.js";
 
 inline XPCShellEnvironment*
-Environment(JS::Handle<JSObject*> global)
+Environment(Handle<JSObject*> global)
```
aeiouaeiouaeiouaeiouaeiouaeiou, line 58: Shouldn't `JS::Handle` be here as it was and as it is in the [current branch](https://github.com/WaterfoxCo/Waterfox/blob/current/ipc/testshell/XPCShellEnvironment.cpp#L57)?

hawkeye116477: Originally (in FF 56 state) it was without assignment to JS namespace, I reverted that together with some other changes, cuz something was messed.

If there are no build issues, then should be fine.
It's only required if there is same reference somewhere else.

PR #2591, opened 2022-02-25
[WFX-346] Add toggle to force light/dark mode in dynamic themes.

### LINKED ISSUE #2562: Can't select Dark Theme when using Lepton Theme on Linux
### What happened?

Using `waterfox-g4-bin-0.6-0`package from AUR on Arch.
### On Linux that doesn't work by any means that I know of.
I only get the Lepton white and can't switch to the dark variant, a thing that on Windows 10 works by nicely asking you which one you want to use just after installation. Is there any way to fix that and make the dark theme work on Linux also? 
I've also tried the official build from the waterfox website but no joy. Thanks

### Reproducible?

- [ ] I have tested this on Firefox.

### Version

G4.𝑋 (Future)

Linux

### Relevant log output

_No response_

adamp01:
Resolve #2562


browser/components/preferences/main.js
```diff
@@ -214,7 +214,7 @@ Preferences.addAll([
   // Media
   { id: "media.hardwaremediakeys.enabled", type: "bool" },
 
-  // Waterfox
```
MrAlex94: Pls have this fixed in the previous PR, makes potential merging/squashing in the future difficult otherwise.

adamp01: The previous PR doesn't touch this file?

MrAlex94 (approved):

PR #2628, opened 2022-03-28
[WFX-371] Increase extensibles test coverage.

adamp01:


browser/extensions/extensibles/content/PrivateTab.jsm
```diff
@@ -400,19 +400,25 @@ const PrivateTab = {
...
-    if (shouldSelect) {
-      let gURLBar = aWindow.gURLBar;
-      let focusUrlbar = gURLBar.focused;
-      gBrowser.selectedTab = newTab;
-      if (focusUrlbar) {
-        gURLBar.focus();
+    try {
+      newTab = gBrowser.duplicateTab(aTab);
+      if (shouldSelect) {
+        let gURLBar = aWindow.gURLBar;
```
davidmaxwaterman: Do we have a policy on destructuring, as is done on 403/404? Ditto line 411. Perhaps there's a complication I'm missing.
Also, I think it's common practice to prefer `const` over `let` - not sure about this code base though.

browser/extensions/extensibles/content/PrivateTab.jsm
```diff
@@ -400,19 +400,25 @@ const PrivateTab = {
...
+        if (focusUrlbar) {
+          gURLBar.focus();
+        }
       }
+      gBrowser.removeTab(aTab);
+    } catch (ex) {
+      // Can use this to pop up failure message
     }
-    gBrowser.removeTab(aTab);
+    return newTab;
```
davidmaxwaterman, line 420: Is returning `undefined` ok? It seems the changes add this possibility.

browser/extensions/extensibles/content/PrivateTab.jsm
adamp01: Line 410 should definitely be `let { gURLBar } = aWindow` at which point it would make sense to simply make line 404 `const { gBrowser, gURLBar } = aWindow`. Based on the Mozilla linter, it also seems that we should no longer be using Hungarian notation - do you think I should have a separate PR to handle changing from Hungarian notation the code that we manage? https://firefox-source-docs.mozilla.org/code-quality/lint/linters/eslint-plugin-mozilla/no-aArgs.html

browser/extensions/extensibles/content/PrivateTab.jsm
adamp01: The return value is only used in some tests as we need to have access to the new tab to be able to remove it at the end of the test. Line 7 in `browser/extensions/extensibles/test/browser/browser_privatetab.js`. I added the try catch block as one of the tests was occasionally failing to find `gBrowser.selectedTab`, if memory serves.
For the general running of the browser the return value would never be used.

browser/extensions/extensibles/content/TabFeatures.jsm
```diff
@@ -172,9 +172,26 @@ const TabFeatures = {
...
+      Ci.nsISupportsPRBool
+    );
+    Services.obs.notifyObservers(
+      cancelQuit,
+      "quit-application-requested",
+      "restart"
+    );
+
+    // Something aborted the quit process.
+    if (cancelQuit.data) {
```
davidmaxwaterman: I dunno, but this seems a little "backward" to me...would it make more sense to `if (!cancelQuit.data) { Services.... }` since it's about to return anyway...or no? (I'm a bit archaic on the whole 'return early' practice).

browser/extensions/extensibles/content/PrivateTab.jsm
davidmaxwaterman: I think a separate PR would be better for merging.

browser/extensions/extensibles/test/browser/browser_privatetab.js
```diff
@@ -1,3 +1,17 @@
...
+requestLongerTimeout(2);
+
+async function togglePrivate(tab, skip = false) {
+  if (skip) {
+    return PrivateTab.togglePrivate(window, tab);
+  }
+  await openTabContextMenu(tab);
+  let openPrivate = document.getElementById("toggleTabPrivateState");
+  openPrivate.click();
+  return gBrowser.selectedTab;
```
davidmaxwaterman, line 12: Question: is the gBrowser.selectedTab updated synchronously? If not, then is this the reason for requesting a longer timeout?

adamp01: No - the longer timeout was actually for the search suggestion block of code, as sometimes it was timing out. Other Moz tests put this at the top of the test file, hence the location of it.

browser/extensions/extensibles/content/TabFeatures.jsm
adamp01: The whole block from 179-194 shouldn't actually be in here, so I will remove it. The `quit-application-requested` observer essentially pops up a confirmation box if certain prefs/conditions are met, this blocks execution so that `if (cancelQuit.data) ` isn't evaluated until the user selects a value from the confirmation box (if it appears). What it ultimately triggers is this method: https://searchfox.org/mozilla-esr91/source/browser/components/BrowserGlue.jsm#2725

For this case we have a separate preference entirely controlling whether a confirmation prompt dispays or not, and display that prompt in the `restartBrowser` method, so adding the `notifyObservers` call is redundant.

browser/extensions/extensibles/test/browser/browser_privatetab.js
davidmaxwaterman: Sorry, I phrased the question poorly. I was wondering if openPrivate.click() causes gBrowser.selectedTab to be updated synchronously - can we be sure it is updated to the new tab as seems to be the intention?

browser/extensions/extensibles/test/browser/browser_privatetab.js
```diff
@@ -27,18 +41,152 @@ add_task(async function testContainer() {
...
+    data => {
+      return (
+        data.state.entries[0].url == URI1 || data.state.entries[0].url == URI2
+      );
+    }
+  );
+  let privateData = closedTabData.filter(data => {
+    return data.state.isPrivate == true;
+  });
+  is(closedTabData.length, 1, "Tab data not stored in private mode");
```
davidmaxwaterman: Is this simply saying 'tab data entry exists for the closed private tab', and the next line is saying, 'yeah, but it's empty'?

browser/extensions/extensibles/test/browser/browser_privatetab.js
```diff
@@ -27,18 +41,152 @@ add_task(async function testContainer() {
...
+  is(privateData.length, 0, "Tab data not stored in private mode");
+});
+
+// Test private->non-private tab is restored on restart
+add_task(async function testRestoreConvertedPrivateTab() {});
```
davidmaxwaterman, line 132: TBD or something?

browser/extensions/extensibles/test/browser/head.js
```diff
@@ -70,6 +128,29 @@ async function openToolbarContextMenu(contextMenu, target) {
   await popupshown;
 }
 
+/**
+ * Helper to paste from clipboard
+ */
+
+async function pasteFromClipboard(browser) {
+  return await SpecialPowers.spawn(browser, [], async () => {
```
davidmaxwaterman: Is async needed here since the function is returning a Promise and there's no awaits?

browser/extensions/extensibles/test/browser/browser_privatetab.js
adamp01: Yeah this is something that we need to be aware of but I haven't yet found a nice way of testing it. It's going to be quite an involved test as we need to create a new private tab 1) directly 2) by converting a non-private tab with the `togglePrivate()` method. After a restart neither of these tabs should be present, but need to have a nice way of triggering a restart and maintaining access to the restarted window object in the test. We also need to do the same thing of creating two private tabs and then converting them to non-private tabs. After a restart both of these should be present as non-private tabs. There was an issue raised a few weeks ago whereby creating a normal tab -> toggling to private -> toggling back to normal would cause the tab to be restored as a private tab.

browser/extensions/extensibles/test/browser/browser_privatetab.js
adamp01: Looking through this the test is maybe a little weird. But my logic was as follows:
1. Open a tab with URI1, toggle it to be private
2. Load URI2 when in private mode so we definitely have some sort of state to save.
3. As toggling private mode duplicates the current tab in the "Private" container and deletes the non-private tab, we should expect `closedTabData` to have one entry, the non-private version of the tab that was remove during toggling. If it has two entries then our private tab state is being stored as well.
4. Private tabs have the `isPrivate` attribute set to true, which should in turn prevent any state being saved. So there should be no state data for tabs with `isPrivate === true`.

browser/extensions/extensibles/test/browser/browser_privatetab.js
adamp01: Ah gotcha, in this case yes - gBrowser.selectedTab is only evaluated after the click is registered.

browser/extensions/extensibles/test/browser/head.js
adamp01: We do need it as we await `SpecialPowers.spawn()`, otherwise we would have a Promise nested within another Promise.

browser/extensions/extensibles/test/browser/browser_privatetab.js
davidmaxwaterman: OK. So my question here is that the two `is()`s have the same comment. I think it might be more intuitive if they were combined, something like:

```
const oneClosedTabWithNoPrivateData = closedTabData.length === 1 && privateData.length === 0;
is(oneClosedTabWithNoPrivateData, true, "Tab data not stored in private mode");
```

or they had different comments as per my interpretation above.

...or we can just leave it if you think it's fine.

WDYT?

browser/extensions/extensibles/test/browser/head.js
davidmaxwaterman: You need `await` on the `SpecialPowers.spawn()`, but not on the callback function? Unless `SpecialPowers.spawn()` somehow expects an async function (not sure how it would tell).
I don't think it matters, tbh, so we can leave this if you like.

browser/extensions/extensibles/test/browser/browser_privatetab.js
adamp01: I like the suggestion you gave - I'll update this now

browser/extensions/extensibles/test/browser/head.js
adamp01: I see what you are saying now, I'll try removing all the async and see what happens

davidmaxwaterman (approved):
lgtm

PR #2631, opened 2022-03-31
[WFX-372] Add preference toggle to enable Geolocation API in Linux distros.

adamp01:
@MrAlex94 - Let me know if you want me to squash any of these together, kept separate so you can squash later.

@davidmaxwaterman - A user on Reddit stated that they were unable to use Hulu on their Linux distro as it was stating that their location could not be determined. After verifying that to be the case using [this](https://developer.mozilla.org/en-US/docs/Web/API/Geolocation_API/Using_the_Geolocation_API#examples), I've taken a look into it. 

It boiled down to Linux distros requiring the Mozilla Location Service as there is no native service that functions correctly. This was previously locked empty, so I have added a toggle on `about:preferences` to enable or disable. The API key should only be added on Linux builds and the toggle should only display on Linux builds, as Windows and Mac do not require this service to function. The keyfile must be in the form of `mozilla-api` and must be added to the mozconfig where we specify it's location, as per [this file](https://searchfox.org/mozilla-esr91/source/build/moz.configure/keyfiles.configure#9).

### DIFF
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -316,6 +316,7 @@ jobs:
       - name: 🏗 Build
         run: |
           ./mach create-mach-environment
+          echo "${{ secrets.MOZ_API_KEY }}" > mozilla-api
           ./mach build
 
       - name: 📦 Package


adamp01:
@MrAlex94 - We need to add a `MOZ_API_KEY` secret to the actions environment as part of this.

Issue #2656, opened 2022-04-06
Translate "Open in new private tab" and more words in options


kamil445:
**Is your feature request related to a problem? Please describe.**
Is there a option to translate this ? Also in Options some words is not translated.

**Describe the solution you'd like**
Update polish translation, translate all untranslated words. I can translate but can't find where i can do that.

**Describe alternatives you've considered**
Another way to translate it manually, via userchrome.css or something...

**Additional context**
screen:
https://i.imgur.com/ayJEeLZ.png

adamp01:
Hi @kamil445 - Want to say thank you for offering to translate, if you take a look at this pull request you can see how translations were done for Japanese, the same principle will apply for Polish.

https://github.com/WaterfoxCo/Waterfox/pull/2615/files

adamp01:
I don't know how comfortable you are doing pull requests etc. so I can always send you a list of things that need translating and can then add it to the browser on your behalf, let me know.

kamil445:
> I don't know how comfortable you are doing pull requests etc. so I can always send you a list of things that need translating and can then add it to the browser on your behalf, let me know.

I updated Translation and my pull request is here https://github.com/WaterfoxCo/Waterfox/pull/2666

I can't find a way to translate this https://i.imgur.com/Jvor2ra.png
In Waterfox Classis this was translated in this file "\browser\chrome\pl\locale\browser\browser.properties"
line "restartPromptQuestion.label".

adamp01:
Perfect - I've got a PR that will be merged shortly that will allow all the translations to be put into the `extensibles.ftl` file. 

I'm actually going to move the restart prompt localisation into `extensibles.ftl` as well. So you could add the translations to `extensibles.ftl` under `restart-prompt-question` and `restart-prompt-title` in preparation for that change.

kamil445:
Thank you, 

I added `restart-prompt-title` and `restart-prompt-question`.
Should i add `%S` like in waterfox classic ?
`restartPromptTitle.label = Restart przeglądarki %S`
`restartPromptQuestion.label = Czy na pewno chcesz ponownie uruchomić przeglądarkę %S?`
or just replace `%S` with `Waterfox ` what i've already done in pull request ?

adamp01:
Ah no - in this case it would be `{ -brand-short-name }`

Thanks for sorting this!

kamil445:
Okey, i changed `Waterfox` to `{ -brand-short-name }`

[...]

PR #2662, opened 2022-04-07
[WFX-378] Simplify Waterfox l10n.

adamp01:
Related to #2656 , aiming to simplify process for translations by moving all translatable items for Waterfox into one fluent file, `extensibles.ftl`.


adamp01:
Still need to remove reliance on `extensibles.properties` and move the restart prompt to use translations from `extensibles.ftl`. We can then just have translators work entirely within extensibles.ftl.

PR #2695, opened 2022-05-03
[WFX-387] [WFX-388] Add translations for several languages.

adamp01:
For reference: including extensibles.ftl directly into browser.jar broke things as soon as localised extensibles.ftl were present. Attempting to put extensibles.ftl into en-US.jar to allow us to have extensibles.ftl in a custom directory caused even more issues, so have resorted to placing the en version in browser/locales/en-US/browser.


MrAlex94:
435bd9a2683e09600655dc322098650479a51a5a
c20964b31ef712a707e0b34342ff8e42d951743c

PR #2773, opened 2022-08-11
[WFX-423] Fix migrateUI error.

adamp01:


waterfox/browser/components/WaterfoxGlue.jsm
```diff
@@ -262,32 +262,40 @@ const WaterfoxGlue = {
...
-      "browser.migration.version"
+      "browser.migration.version",
+      128
     );
 
     async function enableTheme(id) {
-      AddonManager.getAddonByID(id).then(addon => addon.enable());
+      const addon = await AddonManager.getAddonByID(id);
+      // If we found it, enable it.
+      addon?.enable;
```
davidmaxwaterman: This smells fishy...is it really some getter with side-effect? In AddonManager.jsm, it seems to be called as a function that returns a promise.

davidmaxwaterman: Yeah, the previous version called it as a function...though, why it was 'async', I don't know since it doesn't return anything. Perhaps I don't quite 'get' async/await yet :/

davidmaxwaterman (changes requested):

waterfox/browser/components/WaterfoxGlue.jsm
adamp01: Should be called as a function, clipped off the brackets by mistake.

davidmaxwaterman (approved):
lgtm

MrAlex94:
00e8f8af99d402f15a896c09ca58b13211ca3910

PR #2788, opened 2022-09-06
[WFX-455] Update AUS logic, so latest version receives no update and …

MrAlex94:
…previous versions receive latest update.

### DIFF
--- a/.github/workflows/production.yml
+++ b/.github/workflows/production.yml
@@ -56,16 +56,33 @@ jobs:
           fi
       - name: "\U0001F9EA Dry run AUS"
         run: |
-          os=(WINNT_x86_64 Linux_x86_64 Darwin_x86_64-aarch64)
-          for j in ${os[@]}
+          OSA=(WINNT_x86_64 Linux_x86_64 Darwin_x86_64-aarch64)
+          for OS in "${OSA[@]}"
           do
-            update_loc=":s3:aus/update/staging/${{ env.CHANNEL }}/${{ inputs.DISPLAY_VERSION }}/"$j"/update.xml"
-            mkdir -p aus_tmp/$j/
-            rclone copyto :s3:aus/update/staging/${{ env.CHANNEL }}/${{ inputs.DISPLAY_VERSION }}/"$j"/update.xml ./aus_tmp/$j/update.xml
-            sed -i 's/staging/releases/g' ./aus_tmp/$j/update.xml
-            sed -i 's/staging/releases/g' ./aus_tmp/$j/update.xml
-            sed -i 's/staging/releases/g' ./aus_tmp/$j/update.xml
-            rclone moveto ./aus_tmp/$j/update.xml :s3:aus/update/production/${{ env.CHANNEL }}/${{ inputs.DISPLAY_VERSION }}/"$j"/update.xml --dry-run
+            mkdir -p aus_tmp/"$OS"/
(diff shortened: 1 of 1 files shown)


.github/workflows/production.yml
```diff
@@ -59,13 +59,32 @@ jobs:
...
-            update_loc=":s3:aus/update/staging/${{ env.CHANNEL }}/${{ inputs.DISPLAY_VERSION }}/"$j"/update.xml"
-            mkdir -p aus_tmp/$j/
-            rclone copyto :s3:aus/update/staging/${{ env.CHANNEL }}/${{ inputs.DISPLAY_VERSION }}/"$j"/update.xml ./aus_tmp/$j/update.xml
-            sed -i 's/staging/releases/g' ./aus_tmp/$j/update.xml
-            sed -i 's/staging/releases/g' ./aus_tmp/$j/update.xml
-            sed -i 's/staging/releases/g' ./aus_tmp/$j/update.xml
-            rclone moveto ./aus_tmp/$j/update.xml :s3:aus/update/production/${{ env.CHANNEL }}/${{ inputs.DISPLAY_VERSION }}/"$j"/update.xml --dry-run
+            mkdir -p aus_tmp/"$j"/
+            rclone copyto :s3:aus/update/staging/${{ env.CHANNEL }}/${{ inputs.DISPLAY_VERSION }}/"$j"/update.xml ./aus_tmp/"$j"/update.xml
+            sed -i 's/staging/releases/g' ./aus_tmp/"$j"/update.xml
```
adamp01: This seems to just be doing the same thing 3 times?

.github/workflows/production.yml
```diff
@@ -59,13 +59,32 @@ jobs:
           os=(WINNT_x86_64 Linux_x86_64 Darwin_x86_64-aarch64)
           for j in ${os[@]}
```
adamp01: We should make a note to make the variable names logical here, not entirely sure why I ever decided to set this as j

.github/workflows/production.yml
```diff
@@ -59,13 +59,32 @@ jobs:
...
-            sed -i 's/staging/releases/g' ./aus_tmp/$j/update.xml
-            sed -i 's/staging/releases/g' ./aus_tmp/$j/update.xml
-            rclone moveto ./aus_tmp/$j/update.xml :s3:aus/update/production/${{ env.CHANNEL }}/${{ inputs.DISPLAY_VERSION }}/"$j"/update.xml --dry-run
+            mkdir -p aus_tmp/"$j"/
+            rclone copyto :s3:aus/update/staging/${{ env.CHANNEL }}/${{ inputs.DISPLAY_VERSION }}/"$j"/update.xml ./aus_tmp/"$j"/update.xml
+            sed -i 's/staging/releases/g' ./aus_tmp/"$j"/update.xml
+            sed -i 's/staging/releases/g' ./aus_tmp/"$j"/update.xml
+            sed -i 's/staging/releases/g' ./aus_tmp/"$j"/update.xml
+          done
+          for i in $(rclone lsf --dirs-only :s3:aus/update/production/${{ env.CHANNEL }} | awk -v RS= '{$1=$1}1' | sed 's|/||g')
```
adamp01: We should change the `i` over to something logical, like `version` or something. Could be handy to have a comment stating what the `awk | sed` is changing the output from and to.

MrAlex94: Aye, good shout on the comment. Will add now.

PR #2843, opened 2022-10-07
[WFX-503] Add preferences page to customise Lepton.

adamp01:


adamp01:
Need to add a window unload listener to remove all the listeners.

PR #2858, opened 2022-10-12
[WFX-514] Prevent forms from saving history in private tab.

adamp01:
Old fix: https://github.com/WaterfoxCo/Waterfox/pull/2604/files

Have changed the fix slightly to ensure we cover all occasions where a tab should be identifying as private.


MrAlex94 (approved):

feelingwalnut:
@MrAlex94 #2918 #2882 these are still present bugs it seems?

PR #2871, opened 2022-10-18
[WFX-518] Disable Waterfox custom css on non-default themes by default.

adamp01:
Quite a lot going on for this one, and quite a few things to check:
1. Check that with pref at 0, all themes have Lepton applied.
2. Check that at 1, only the Moz themes have Lepton applied.
3. Check that at 2, no themes have Lepton applied.
4. Check flipping the pref from 2 to 1 or 1 to 2 on a Moz theme applies/removes Lepton.
5. Check that flipping the pref from 1/2 to 0 or 0 to 1/2 on a non Moz theme applies/removes Lepton.
6. Check that on `about:preferences#theme` that disabling Waterfox customisations hides all the pref toggles.
7. Check that enabling customisations shows the toggles.
8. Check that with pref at 0 that Waterfox css is applied at startup.
9. Check that `userChrome.theme.enable` is migrated successfully only on first launch. Subsequent launches should not override user set value.


davidmaxwaterman (approved):
I can't see anything wrong.

davidmaxwaterman:
I successfully confirmed checks #1-8, the only remaining issue is #9 which Adam is going to check out.

davidmaxwaterman:
ok step 9 now working

davidmaxwaterman (approved):
all checks pass and code looks good to me

MrAlex94 (approved):

PR #2875, opened 2022-10-20
[WFX-524] Look for NativeMessagingHosts in Mozilla directory.

adamp01:
Should we replace the `XRESysNativeManifests` as well?


MrAlex94 (changes requested):
We should override `XRESysNativeManifests` as well, as you've mentioned.

adamp01:
Have updated now.

MrAlex94 (approved):

Issue #2924, opened 2022-12-10
Consider building in JPEG-XL support


jobbautista9:
Currently, there's an `about:config` option for JPEG-XL support, but it doesn't do anything at the moment. It would be nice if Waterfox adds in the `--enable-jxl` option in the standard mozconfig. The decoding works fine, though alpha, animations, HDR, color profiles, and progressive decoding aren't supported yet (all of which except HDR have pending patches in Mozilla phabricator which we could apply ourselves in our fork btw, see the depends of [bug 1539075](https://bugzilla.mozilla.org/show_bug.cgi?id=1539075)). AVIF was shipped without support for animations either, so why should JPEG-XL be different?

I think having JPEG-XL support out-of-the-box would be a nice selling point of the browser, and help Waterfox differentiate itself from its parent.

MrAlex94:
Done! https://github.com/WaterfoxCo/Waterfox/commit/4e1dca751406fe4a71435f2088c9ac4eae7d41ad

jobbautista9:
Thank you! :D

louie-github:
Does this actually enable JXL out-of-the-box? Looking at [moz.configure](https://github.com/WaterfoxCo/Waterfox/blob/4e1dca751406fe4a71435f2088c9ac4eae7d41ad/toolkit/moz.configure#L692), there may be a Nightly check that you have to disable:

```python
@depends("--disable-jxl", milestone.is_nightly)
def jxl(value, is_nightly):
    if is_nightly and value:
        return True
```

Nulled-Out:
> Does this actually enable JXL out-of-the-box? Looking at [moz.configure](https://github.com/WaterfoxCo/Waterfox/blob/4e1dca751406fe4a71435f2088c9ac4eae7d41ad/toolkit/moz.configure#L692), there may be a Nightly check that you have to disable:
> 
> ```python
> @depends("--disable-jxl", milestone.is_nightly)
> def jxl(value, is_nightly):
>     if is_nightly and value:
>         return True
> ```

So, to just set it as false and submit in a new PR?

RobertZenz:
As a heads up, enabling JPEG XL support seems to break various websites in G5.1.1, for example [MagicHolz](https://magicholz.de/).

[...]

PR #2936, opened 2022-12-21
Disable Nightly check for JPEG XL support

louie-github:
As mentioned in #2924, 4e1dca751406fe4a71435f2088c9ac4eae7d41ad does not actually affect anything, because JPEG XL support is disabled in anything but Nightly builds.

Weirdly, the `image.jxl.enabled` config option is already set to `true` by default in [000-waterfox.js](https://github.com/WaterfoxCo/Waterfox/blob/4e1dca751406fe4a71435f2088c9ac4eae7d41ad/waterfox/browser/app/profile/000-waterfox.js#L98). This means that **all Waterfox releases** have broken JPEG XL support, since Firefox (and therefore Waterfox) sends the `Accept: image/jxl` header even if the JXL decoding support is not actually built and present.

This leads to issues such as [missing pictures on Shopify-powered sites](https://github.com/WaterfoxCo/Waterfox/issues/2924#issuecomment-1361054106):
> As a heads up, enabling JPEG XL support seems to break various websites in G5.1.1, for example [MagicHolz](https://magicholz.de/). With enabled JPEG XL support (image.jxl.enabled set to true) various images on the website do not load anymore.

This should fix it by _actually_ enabling JPEG XL support out-of-the box in all builds by removing the Nightly check.

Additionally, this test page by Jon Sneyers may prove useful: https://jpegxl.info/test-page/

### DIFF
--- a/toolkit/moz.configure
+++ b/toolkit/moz.configure
@@ -689,9 +689,9 @@ set_define("MOZ_AV1", av1)
 option("--disable-jxl", help="Disable jxl image support")
 
 
-@depends("--disable-jxl", milestone.is_nightly)
-def jxl(value, is_nightly):
-    if is_nightly and value:
+@depends("--disable-jxl")
+def jxl(value):
+    if value:
         return True
 
 


MrAlex94:
Thanks!

PR #3033, opened 2023-05-11
Add a few missing <cstdint> includes

hawkeye116477:
> GCC 13s libstdc++ reduced its dependency on some headers like <cstdint>, so it's no longer transitively included through various headers. Include it explicitly.

> See also: https://gcc.gnu.org/gcc-13/porting_to.html#header-dep-changes
> 
>   11:16.32 .../components/telemetry/pingsender/pingsender.cpp:30:7:
>                     error: unknown type name 'uint32_t'
>   11:16.33 const uint32_t kConnectionTimeoutMs = 30 * 1000;
>   11:16.33       ^
>   11:16.33 ... /components/telemetry/pingsender/pingsender.cpp:76:9:
>                     error: unknown type name 'uint32_t'
>   11:16.33   const uint32_t kBufferSize = 4 * 1024 - 1;
>   11:16.34         ^
>   11:16.44 2 errors generated.
>   (et al)

### DIFF
--- a/dom/media/webrtc/sdp/RsdparsaSdpGlue.cpp
+++ b/dom/media/webrtc/sdp/RsdparsaSdpGlue.cpp
@@ -4,6 +4,7 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this file,
  * You can obtain one at http://mozilla.org/MPL/2.0/. */
 #include <string>
+#include <cstdint>
 
 #include "sdp/RsdparsaSdpInc.h"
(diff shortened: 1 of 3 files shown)


hawkeye116477:
As I see you reverted it now for some reason.

MrAlex94:
That’s odd - I’ve only done a rebase, shouldn’t have gotten rid of the commit. I’ll add it back it! Sorry about that.

MrAlex94:
Has been fixed!

PR #3079, opened 2023-07-20
Update license field following SPDX 2.1 license expression standard

frisoft:
The new recommendation is to follow the SPDX 2.1 standard. This allows automatic license verification software to work properly. Reference: https://doc.rust-lang.org/cargo/reference/manifest.html#the-license-and-license-file-fields


MrAlex94:
Thanks for this - would you consider opening a bug in [Bugzilla](https://bugzilla.mozilla.org/) for this? Otherwise I feel it may cause merge issues in the future, and better for it to be upstream.

If Mozilla won't accept it, might be worth pulling in then.

Issue #2962, opened 2023-01-19
G5.1.12 automatically downloads JPEG-XL when I try to open image in a new tab


thekcool:
### What happened?

A bug happened!

Posted on reddit https://www.reddit.com/r/waterfox/comments/10flkpn/g512_automatically_downloads_jpegxl_when_i_try_to/

Just upgraded to G5.1.2 and the release page says "JPEG-XL is now enabled by default." so I clicked the link (https://jpegxl.info/test-page/) to try it out. When I right click on the JPEG-XL image & "Open image in new tab", WF downloads the image & opens in new tab https://jpegxl.info/test-page/red-room.jxl (which is blank) & then opens the downloaded image in a new tab. The other images (.webp, .jpg, .png) open in a new tab as expected.

So WF

1. downloads the .jxl
2. opens a new tab with a blank page
3. opens the downloaded .jxl in a new tab

### Reproducible?

- [ ] I have tested this on Firefox.

### Version

G5

### What platform are you seeing the problem on?

Windows

dfgrgavr:
Hi! 
Same problem, as i tested, its in the Firefox109 too, so maybe Waterfox cant do much, until Firefox do.

PR #3262, opened 2023-11-09
Extend JPEG XL detection/recognition (also AVIF a tiny bit)

goodusername123:
Adds JPEG XL recognition to a few missing spots, Should hopefully fix issue #2962 however I can't test myself since I don't have a setup to compile Waterfox builds of my own.

### DIFF
--- a/browser/components/BrowserContentHandler.sys.mjs
+++ b/browser/components/BrowserContentHandler.sys.mjs
@@ -1018,6 +1018,7 @@ function maybeRecordToHandleTelemetry(uri, isLaunch) {
       ".xhtml",
       ".svg",
       ".webp",
+      ".jxl",
     ]);
     if (registeredExtensions.has(extension)) {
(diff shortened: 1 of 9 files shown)


MrAlex94:
Thanks for submitting. Will take a look and revert.

goodusername123:
@MrAlex94 
After doing another pass over the codebase I have come to realize there are quite a few spots that I have missed, Would it be fine if I attached another commit to this pull request with the extra places I originally missed? or would you prefer a separate pull request? The commit would be larger then the other ones currently present and would also include adding JXL as a supported file association type under Windows, along with some minor spots where AVIF recognition is also missing.

MrAlex94:
You can push it here, no problem. I typically squash down commits to keep it manageable for rebasing on top of gecko-dev

MrAlex94:
Thanks for the updated commits. Will revert ASAP.

PR #3294, opened 2023-12-10
Windows: Restore support for native styled scrollbars. (external patch)

goodusername123:
Patch is sourced from: https://github.com/ephemeralViolette/firefox-native-controls
documentation/background for the patch is here: https://github.com/ephemeralViolette/firefox-native-controls/blob/main/docs/scrollbars.md

This will allow Waterfox to use natively styled scrollbars on Windows when `widget.non-native-theme.enabled` is changed from it's default value of "true" to "false" (this requires a restart of the browser to function)

This will solve https://github.com/WaterfoxCo/Waterfox/issues/3229

### DIFF
--- a/widget/windows/nsNativeThemeWin.cpp
+++ b/widget/windows/nsNativeThemeWin.cpp
@@ -40,6 +40,7 @@
 #include "nsWindow.h"
 #include "prinrval.h"
 #include "WinUtils.h"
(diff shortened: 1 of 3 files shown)


IS4Code:
What a delightful Christmas gift, just a month later after I started using Waterfox again (I was stuck at Firefox 96 previously due to the scrollbars). Thanks!

PR #3369, opened 2024-03-02
Selector improvement and code cleanup

FaySmash:
Improved the download button selector and removed unused and non-functional code (still WIP)

The old code only worked on the English version of the Chrome Web Store because it searched for "Add to Chrome", which changes with different localization. I made the selector generic to work on all languages. I also found that the function `parentNodeUntil` doesn't get called and the `browser.runtime.onMessage` and `browser.runtime.sendMessage` don't work and also don't have any purpose (?).
The installation still doesn't work tho, also the `uninit` function doesn't get called so this PR isn't final but servers as a base for further improvements.


MrAlex94:
Sorry for the mess in commits - if your rebase your change ontop of the latest HEAD, should get cleaned up (may require a force push to your branch).

I tested this, but when en-US remains the default, the button detection no longer works at all.

Without patch:
image

With patch:
image

FaySmash:
> Sorry for the mess in commits - if your rebase your change ontop of the latest HEAD, should get cleaned up (may require a force push to your branch).
> 
> I tested this, but when en-US remains the default, the button detection no longer works at all.


Hi Alex, thanks for getting back at me. I changed the JS Selector to identify the button to this: `(document.querySelectorAll('main > div > section > section > div > div > button'))[0]`

Which works in my case:
Screenshot 2024-03-20 185504
(also tested on other browsers)

As I noted, this PR is not final and still needs work (currently even if the button appears, the installation still doesn't work)

Kichura:
Shouldn't this be a draft in terms of WIP context?

PR #3406, opened 2024-04-03
Change Oblivious DNS over HTTP to Oblivious DNS over HTTPS

Dyras:
Hello! 

For some reason there is no S in Oblivious DNS over HTTPS almost anywhere in this project.

For example:
https://www.waterfox.net/blog/waterfox-in-2023/#dns-over-oblivious-http
https://www.waterfox.net/docs/releases/G6.0.1/

More importantly, it says Oblivious DNS over HTTP in the browser itself:
image

Is the implementation in Waterfox unencrypted, or is the missing S a typo?

Regardless, this fixes that in the descriptions/comments and in the en-US locale of the browser.

There are still functions named stuff like oblivious_http but I ain't touching that.

In an unrelated note, it seems impossible to use anything but CloudFlare for this feature. Any plans of changing that? Just curious.

Thanks for making this browser!

### DIFF
--- a/netwerk/dns/TRRService.cpp
+++ b/netwerk/dns/TRRService.cpp
@@ -55,7 +55,7 @@ constexpr nsLiteralCString kTRRDomains[3][7] = {
     "private.canadianshield.cira.ca"_ns,
     "doh.xfinity.com"_ns,  // Steered clients
     "dns.shaw.ca"_ns, // Steered clients
-    "dooh.cloudflare-dns.com"_ns, // DNS over Oblivious HTTP
+    "dooh.cloudflare-dns.com"_ns, // DNS over Oblivious HTTPS
     },
     {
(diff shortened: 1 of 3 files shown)


MrAlex94:
Thanks for taking the time to submit a PR! Oblivious HTTP is the correct usage of the term as you can see in the [IETF proposal](https://ietf-wg-ohai.github.io/oblivious-http/draft-ietf-ohai-ohttp.html).

The reason being:

> Focus on Application Layer:  OHTTP primarily aims to obfuscate information at the HTTP request and response level, not the underlying transport layer.  HTTPS, on the other hand, secures the connection itself. OHTTP builds on top of HTTPS for the secure transport, but its core function lies in making the HTTP request and response data invisible to intermediaries.

Dyras:
Well I be damned!

PR #3451, opened 2024-05-21
Fix building with Rust 1.78

hawkeye116477:


MrAlex94:
I ported down the changes before, but it was causing crashes on some systems - I'll see if this way work's on those systems.

hawkeye116477:
Looks like Mozilla approved patch for Bug 1882291 for ESR. However not sure if that's enough, will check later again.

PR #3487, opened 2024-06-26
Fix geolocation api setting always showing as enabled

### LINKED ISSUE #3441: Geolocation API always turns on
### What happened?

I prefer to disable the Geolocation API, but everytime I start Waterfax it's enabled again.
Is it only me experiencing this issue?
Version G6.0.13

### Reproducible?

- [ ] I have checked that this issue cannot be reproduced on Mozilla Firefox.

### Version

G6

Linux

### Relevant log output

_No response_

LAURAilway:
The radio buttons didn't correctly detect "" as a value, so the geolocation setting always appeared as "enabled" on reload, regardless of actual state. Setting the string for "disabled" to " " seems to fix the issue. 
I am not entirely sure if this is bad practice for whatever reason, I defer that judgement to you.
Closes #3441


LAURAilway:
Note: Setting it to `about:blank` (or any other non-empty string that isn't a valid url) also works, if that would be preferable for easier maintenance

MrAlex94:
https://github.com/BrowserWorks/Waterfox/commit/4a07e9784069ce21ba99876774753e4e8d605c77

PR #3540, opened 2024-09-11
Fix appears even when the data-category at turn on

black7375:
## Descriptions

1. Prevent the elements of `#tabsSidebar` from being displayed when the browser is first opened.
2. Overall order adjustment and preset changes for the Look & Feel page
   - Designed with a focus on purpose rather than functionality
   - Use a structure similar to other Firefox settings pages

## Screenshots
image
image
image

## Additional context
- Blocked-by: https://github.com/BrowserWorks/l10n/pull/6


black7375:
@MrAlex94 I think i18n should also be merged/rebased because of some added words.
https://github.com/BrowserWorks/l10n/pull/6

PR #3554, opened 2024-09-30
Trust additional binaries signed by Microsoft

goodusername123:
This fixes about:third-party not trusting some DLLs from Microsoft that are required similar to https://github.com/BrowserWorks/Waterfox/pull/3269.
makes it so DLLs such as `vcruntime140.dll`, `vcruntime140_1.dll`, and `msvcp140.dll` stop being detected in about:third-party, The reason this happens in the first place is because it seems like only the first signature string within a DLL is checked and for these DLLs "`Microsoft Windows Software Compatibility Publisher`" is the first signature.


goodusername123:
Well this is a bit embarrassing but I didn't realize that the main/current branch was switched out with Waterfox 6.5.X and this pull request was meant to solve an issue that *might* only be present in Waterfox G6, I haven't tried out the 6.5 beta but a check of the installer revealed that the mentioned DLL files are no longer shipped with Waterfox and are instead grabbed/called from system32 (which is a good thing!) meaning that this pull request probably isn't needed/relevant anymore since as far as I understand `about:third-party` automatically trusts DLLs present in system32.
Feel free to close/not merge this pull request because it probably isn't needed anymore (I might just close it myself).

PR #3760, opened 2025-05-11
style: update Waterfox icon to match Apple's macOS design guidelines

rymayerhofer:
Reduced the size of the Waterfox logo to ring 3 of the MacOS icon template, added a white background with a linear gradient, and added a drop shadow to the logo. This makes the application icon much more consistent with Apple's macOS design guidelines as defined [here](https://developer.apple.com/design/human-interface-guidelines/app-icons/).


MrAlex94:
Sorry for the delay on getting back to you!

For v6.6 when I updated the icon to support Liquid Glass, I went ahead and updated the icon to match.

PR #4048, opened 2026-01-11
(Take 2) Attempt to restore Ogg Theora video support

goodusername123:
The same as https://github.com/BrowserWorks/waterfox/pull/3975 but with the following change(s):
- Updated to [Theora 1.2](https://xiph.org/press/2025/theora-release-1.2/) (used to be a alpha version of 1.2 before this).
- ~~Set to off/disabled by default~~ (Not pushed because I wasn't 100% sure if this was a requirement of yours for inclusion into Waterfox based on your comment/response on the previous PR. If desired/required I can push a commit to flip this off by default).
- Didn't include the file "`clang-arm.patch`" since it's pretty unnecessary and is also leftover from before the changes within it where merged into libtheora officially.
- Pushed very small minor lines/changes that GitHub failed to upload before (the commit with "followup" in it's name).
Changes added after submitting this PR:
- Added a extra copyright line to license.html (also corrected accidentally removing info for something else however this wasn't present as an issue in the original PR).
- Added missing Solaris OS define for the RDD pref.

Also here's one of the test files from the previous PR for easy testing: https://threejs.org/examples/textures/sintel.ogv


toolkit/content/license.html
```diff
@@ -2428,9 +2428,6 @@ <h1><a id="chromium"></a>Chromium License</h1>
                 <li><code>toolkit/components/formautofill/shared/FormAutofillNameUtils.sys.mjs</code></li>
                 <li><code>editor/libeditor/EditorEventListener.cpp</code></li>
                 <li><code>security/sandbox/</code></li>
-#ifdef MOZ_ZUCCHINI
```
MrAlex94, line 2431: This shouldn't be removed

toolkit/content/license.html
```diff
@@ -5316,9 +5313,6 @@ <h1><a id="unicode"></a>Unicode License</h1>
               <li><code>intl/icu</code></li>
               <li><code>intl/tzdata</code></li>
               <li><code>js/src/util</code></li>
-#ifdef MOZ_ZUCCHINI
```
MrAlex94, line 5319: +1

modules/libpref/init/StaticPrefList.yaml
```diff
@@ -11420,6 +11420,23 @@
   mirror: always
 #endif
 
+- name: media.rdd-theora.enabled
```
MrAlex94, line 11423: Realistically, this should be false for all (and no need for so many if guards since the values are all the same.

MrAlex94 (changes requested):

toolkit/content/license.html
goodusername123: > This shouldn't be removed

Ah whoops sorry about that, my local copy of the files/changes for this branch was from before this info was present so when I pushed it removed this by mistake.

modules/libpref/init/StaticPrefList.yaml
goodusername123: This is seemingly the way it is due to this being intended to only be enabled for platforms that fully support the multi-process architecture of Gecko. It manually sets each and every desktop platform to true and then gets set to false for mobile and esoteric/unknown platforms. This matches how it's setup for most of the other RDD process related prefs for similar formats, The only thing missing/different compared to the others is the define for Solaris OS which was added after Theora was removed. ill push a commit to add the missing Solaris define for the niche scenario of Waterfox getting compiled for it (unofficially or officially). Also I do agree that this here looks a bit messy but changing it would put the logic for this pref out of sync with all the other related/similar prefs, I mean I guess the same decision logic could be kept by just having converting a bunch of the if checks to all that place at once but this would still break consistency with the other prefs.

Oh also just to be clear this isn't the pref for enabling or disabling Theora support outright that would be `media.theora.enabled` which is somewhere lower down in the file.

MrAlex94 (approved):

